PR #1, opened 2015-05-02
Use the `rvm` key to specify Ruby versions

BanzaiMan:
See http://docs.travis-ci.com/user/languages/ruby/#Choosing-Ruby-versions-and-implementations-to-test-against


BanzaiMan:
Looks like 2.1.6 has a problem loading the dependencies correctly (`Digest`, maybe, but it could be something else).

PR #4, opened 2015-05-06
Support for aws-sdk 2

peeyush1234:
I would like to get some feedback on the pull request for using aws-sdk 2. Once this pull request is accepted, the dynamoid major version should be bumped as the change is not compatible with aws-sdk 1.

In the meantime (till, I get the feedback on the pull request), I will update the README section to reflect the changes.


BilalBudhani:
This is great @peeyush1234. I will try this in my current project and update here.

peeyush1234:
@BilalBudhani, any feedback will be appreciated.

BilalBudhani:
I've tested this in my current project. Everything seems to be working fine at the moment. I just ran into one breaking issue which is when the field is not defined in the model but it exits in DynamoDB item. This behaviour should be handled as there are chances of DynamoDB fields being out of sync with the model. It might break some legacy applications.

philipmw:
@BilalBudhani, how do you suggest to handle that situation? One problem I see is that without a `field` declaration, Dynamoid won't know how to materialize the string value.

BilalBudhani:
@philipmw Can we just simply ignore the `fields` which is not declared in the model? This way Dynamoid will only provide access to `fields` which has been declared. Make sense?

philipmw:
@BilalBudhani, I must have misunderstood your original bug report. I agree that Dynamoid should ignore fields that aren't declared on the model. It sounds like you're saying that if a DynamoDB record has attributes that aren't declared on the Ruby model, Dynamoid raises an exception when materializing the model instance. Is that right?

BilalBudhani:
@philipmw Yes. When DynamoDB record has a `field` which is not declared on Dynamoid model. It raises an exception.

philipmw:
@BilalBudhani would you open an issue about this? I assume this behavior exists on the master branch too.

BilalBudhani:
@philipmw Sure. I will open an issue for it.

loganb:
@philipmw, can you move your commits that remove functionality and/or fix issues other than AWS-SDK-2 to a separate PR?

philipmw:
@loganb, done.

PR #16, opened 2015-07-14
make find return array once array argument passed in

mengwan:
The change I made is to make the gem's finders::finder behave more like an active::record find. In particular, I made so that find(["#id"]) return [Object] instead of just Object


peeyush1234:
Change looks good to me. I will wait for Philip's comments before merging the pull request.

philipmw:
Looks good. Thanks.

PR #17, opened 2015-07-15
Remove support for indexes

philipmw:


langalex:
what was the reasoning behind removing indexes? what's the recommended way to use dynamodb's native indexes?

philipmw:
Hello, @langalex. We removed the current implementation of indexes because it was a hack necessitated by a lack of true index support on DynamoDB's side. Today, you can create native indexes in DynamoDB, but the latest version of Dynamoid won't know how to take advantage of them. (It'll SCAN instead of QUERY.)

We thought it's important to remove the current implementation before releasing the next major version, so that the next implementation does not have to be backwards-compatible.

Re-adding support for indexes (without the drag of backward compatibility) is open to whomever needs them first.

langalex:
thanks that answers my question. i might take a stab at it.

dishcandanty:
Any updates on indexes being added back in? Its a bit scary to get a bunch of
```
Queries without an index are forced to use scan and are generally much slower than indexed queries!
```

And not have anything listed in the current documentation, or any way to suppress or actually resolve the issues :/

philipmw:
Hi, @dishcandanty. I am sorry to say, as far as I know, no one is working on adding proper support for indexes right now.

dishcandanty:
Maybe not the right place for this question. But how does that effect using this gem in a production situation? Seems like without Indexes performance will be particularly hampered? 

For example, I imagine this would be using the indexes, and is popping up the indexes warning:
```
User.where(:state => 'active').all
```

sumocoder:
I actually did some work on this. I'll send a pull request today.

PR #21, opened 2015-07-16
Field class type

philipmw:


README.markdown
```diff
@@ -119,6 +124,57 @@ You can optionally set a default value on a field using either a plain value or
...
+  class MoneyAdapter
+    def self.dynamoid_load(money_serialized_str)
+      Money.new(...)
+    end
+    
+    def self.dynamoid_dump(money_obj)
+      money_obj.value.to_s
+    end
+  end
+```
```
loganb: Include in this code example how you declare the field to use the MoneyAdapter

spec/dynamoid/persistence_spec.rb
```diff
@@ -313,4 +313,90 @@
...
+        end
+      end
+
+      before(:each) do
+        subject.price = Money.new(BigDecimal.new('5'))
+        subject.save!
+        subject.reload
+      end
+
+      it 'round-trips using Dynamoid-specific methods' do
```
loganb: This block of assertions looks about the same across the contexts. Would a shared example group DRY this up a bit?

README.markdown
philipmw: done.

spec/dynamoid/persistence_spec.rb
philipmw: There are enough differences that I don't see a good way of factoring out the commonality. These three scenarios share only one spec: "it 'round-trips using Dynamoid-specific methods'"... I don't want to make a shared example block for just one spec.

Issue #10, opened 2015-07-02
`Dynamoid::Adapter.reconnect!` is not thread-safe


philipmw:
A backtrace when a program uses Dynamoid (with AWS SDK v2) in a multi-threaded environment:

     NoMethodError:
       undefined method `list_tables' for nil:NilClass
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter/aws_sdk_v2.rb:220:in `list_tables'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:128:in `list_tables'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:42:in `benchmark'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:128:in `list_tables'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:28:in `reconnect!'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:42:in `benchmark'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:28:in `reconnect!'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:17:in `adapter'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:128:in `get_item'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:42:in `benchmark'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:128:in `get_item'
     # .../dynamoid-1.0.0pre/lib/dynamoid/adapter.rb:80:in `read'

Here's what I think is happening:
  1. Multiple threads simultaneously call `Dynamoid::Adapter.get_item`.
  2. Because Dynamoid isn't connected to a DynamoDB instance yet, multiple threads invoke `Dynamoid::Adapter.reconnect!`. Suppose thread A and B are at the start of `.reconnect!`.
  3. Thread A instantiates an adapter, assigns it to `@adapter`, and calls `@adapter.connect!` (lines 26--27 of adapter.rb). _Context switch to thread B._
  4. Thread B instantiates an adapter, assigns it to `@adapter`, but does not yet call `.connect!`. _Context switch to thread A._
  5. Thread A calls `list_tables` (line 28 of adapter.rb) on an adapter instance that hasn't connected yet.

philipmw:
[...]

PR #23, opened 2015-07-16
Fix threadsafety of Dynamoid::Adapter (issue #10)

philipmw:
I changed Dynamoid::Adapter from being a module having global state to a class
that's instantiated by Dynamoid module.  This makes Adapter more testable and
allows me to write specs verifying its lazy-connection behavior.

I extend the lazy behavior to caching tables -- instead of caching tables when
establishing a connection, we cache on first demand.

Turning Dynamoid::Adapter into a class required me to rename the `Adapter`
namespace (module) of AwsSdkV2 to `AdapterPlugin`.


philipmw:
@loganb Can I get your eyes on it?

lib/dynamoid/adapter.rb
```diff
@@ -5,27 +5,48 @@ module Dynamoid
...
+      @tables_mutex = Mutex.new
     end
 
-    # The actual adapter currently in use: presently AwsSdk.
-    #
-    # @since 0.2.0
-    def self.adapter
-      reconnect! unless @adapter
-      @adapter
+    def tables
```
loganb: So the double-check locking idiom is (mostly) not safe. In very rare circumstances, I think this code can fail. See: http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -1,5 +1,5 @@
 module Dynamoid
-  module Adapter
+  module AdapterPlugin
```
loganb: So how does this work given that the class name doesn't match the file name?

loganb:
I like the move away from global state and towards instantiated classes. Do you think you can avoid locking entirely by eagerly instantiating the client and assigning it to a variable? If not, you may want to consider using an atomic variable rather than a mutex, or just always lock when retrieving the adapter.

lib/dynamoid/adapter.rb
philipmw: Ok, I will switch to something else.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
philipmw: They do match... the module name is `AdapterPlugin` while its directory is `adapter_plugin`, and the class name is unchanged.

philipmw:
Ok, I'll see what I can do. Thanks for the feedback.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
loganb: Whoops! My bad. :)

philipmw:
This is ready for another review from someone.

philipmw:
This got auto-closed because I merged (and subsequently deleted) a branch atop which this PR is based.

PR #29, opened 2015-12-18
Added support for optimistic locking on delete

sumocoder:


sumocoder:
Please ignore the empty line changes, my editor is just stripping out trailing whitespaces. I did not make any changes to the `destroy_all` method since that does batch writes and batch writes do not support conditional checks.

sumocoder:
I have also noticed that using integers for lock version is not ideal since you can bypass the lock version check if you put in a decremented value. For example the following spec fails

```ruby
it 'prevents concurrent writes even when lock version is dirty (modified)' do
    address.save!
    a1 = address
    a2 = Address.find(address.id)

    a1.lock_version = 0
    a1.city = 'Seattle'
    a2.city = 'San Francisco'

    a1.save!
    expect { a2.save! }.to raise_exception(Dynamoid::Errors::StaleObjectError)
end
```

It might be better to use SecureRandom.hex to randomly generate a lock_version string instead of using integers and incrementing them

lib/dynamoid/persistence.rb
```diff
@@ -216,7 +216,21 @@ def destroy
...
+          if changes[:lock_version].nil?
+            self.lock_version
+          else
+            changes[:lock_version][0]
+          end
+        options[:conditions] = conditions
+      end
       Dynamoid.adapter.delete(self.class.table_name, self.hash_key, options)
+    rescue Dynamoid::Errors::ConditionalCheckFailedException
+      raise Dynamoid::Errors::StaleObjectError.new(self, 'delete')
```
philipmw: Can you make this a symbol, for GC-friendliness?

spec/dynamoid/persistence_spec.rb
```diff
@@ -292,6 +292,32 @@
         msg.destroy
       end.to_not raise_error
     end
+
+    context 'with lock version' do
+      it 'deletes a record if lock version matches' do
+        address.save!
+        expect { address.destroy }.to_not raise_error
```
philipmw: If you're expecting an exception, don't you have to use `#destroy!` instead of the kinder `#destroy`?

spec/dynamoid/persistence_spec.rb
```diff
@@ -292,6 +292,32 @@
...
+
+      it 'does not delete a record if lock version does not match' do
+        address.save!
+        a1 = address
+        a2 = Address.find(address.id)
+
+        a1.city = 'Seattle'
+        a1.save!
+
+        expect { a2.destroy }.to raise_exception(Dynamoid::Errors::StaleObjectError)
```
philipmw: same

spec/dynamoid/persistence_spec.rb
```diff
@@ -292,6 +292,32 @@
...
+      end
+
+      it 'uses the correct lock_version even if it is modified' do
+        address.save!
+        a1 = address
+        a1.lock_version = 100
+
+        expect { a1.destroy }.to_not raise_error
```
philipmw: same

philipmw:
Thanks for the PR! Looks great overall. I just added some small stuff. I'll think about what you said regarding using a random string.

spec/dynamoid/persistence_spec.rb
sumocoder: The destroy method in ActiveRecord does not catch ActiveRecord::StaleObjectError, so its the responsibility of the caller to handle those exceptions.

https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activerecord/lib/active_record/persistence.rb#L179

I tried to keep the same format as ActiveRecord. Although I understand what you mean, as a reader it looks like destroy does not throw an exception, I can make the modification.

lib/dynamoid/persistence.rb
sumocoder: Sure, would you like me to convert it to string for attempted_action attribute on the exception. It would be a breaking change otherwise.

philipmw: No I see that ActiveRecord also uses strings here.

PR #32, opened 2016-01-11
Added support for hash, binary and sets

sumocoder:
This is just for initial review, I will add more specs if you guys are ok with the changes. 

I have also added additional type checking to ensure that only correct types are allowed to save. Eg: If you pass an array to an integer field, it will save nil, this is similar to how active record does it.

Here is a summary of the mappings

Dynamoid type | Ruby type | AWS DynamoDB type
-----------------|-----------------|------------------
integer | Integer | Number
number | BigDecimal | Number
string | String | String
datetime | DateTime | Number
boolean | true/false | Boolean
serialized | String | String
array | Array | List
hash | Hash | Map
set | Set | Type is based on first element in the set
string_set | Set of Strings | String Set
number_set | Set of BigDecimal | Number Set
binary_set | Set of StringIO | Binary Set

I think calling undump when an object is initialized is not needed and adds unnecessary complexity. It should only be called when an item is loaded from database. The aws-sdk gem already does a lot of the unmarshalling on its end. This could be a breaking change though.

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -404,7 +404,7 @@ module Dynamoid
 
       STRING_TYPE  = "S".freeze
       NUM_TYPE     = "N".freeze
-      BOOLEAN_TYPE = "B".freeze
+      BINARY_TYPE = "B".freeze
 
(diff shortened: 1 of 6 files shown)


sumocoder:
Not sure why this is failing on jruby-head

PR #33, opened 2016-01-11
Added fix to ignore unknown fields on model initialize

sumocoder:


philipmw:
Thanks. This looks great! JRuby is broken, but it doesn't appear to be related to your PR, so I'll accept it and investigate the failure. My guess is that a newer version of the Ruby AWS SDK changed some things on us.

PR #35, opened 2016-01-28
fix query attribute behavior for booleans

amirmanji:
The current behavior of query attributes seems unexpected for booleans set to false:

```ruby
address.deliverable = false
address.deliverable?
=> true # because deliverable is not nil
```

This pull request changes the behavior to match ActiveRecord's:

```ruby
address.deliverable = false
address.deliverable?
=> false
```


philipmw:
Thanks!

PR #38, opened 2016-01-29
Added destroy! and delete_table

sumocoder:


philipmw:
Great. Would you please rebase?

pboling:
@sumocoder Finally fixed the spec suite on master.  I also added a synchronous option to the adapter's delete_table.  Do you think this should use that (basically waiting to return from the method until the delete is complete - blocking), or should it be an option here?

Is there any chance you could rebase?

pboling:

Issue #5, opened 2015-06-24
Provide option to skip namespace in configuration


BilalBudhani:
I've been rewriting a legacy app in Ruby. When I try to provide a custom name to my table name Dynamoid prepends namespace to the table name. It would be great if we can skip namespace in configuration. I also tried leaving the namespace blank which results in `_` before the table name.

RedaBenh:
+1
i try to name my table "my_table". but i can't because the gem set it to "_my_table" !

cignoir:
+1
I hope this gets fixed.

philipmw:
I merged the fix, thanks to brenden.

PR #40, opened 2016-01-31
Enable creating tables with no namespace prefix

brenden:
This PR allows table namespacing to be avoided when the namespace config option is set to nil. It fixes issue #5.


cignoir:
I need this and hope to be fixed #5 too.

philipmw:
Thanks!

PR #47, opened 2016-02-25
Fix the single association create method. It should not raise an exce…

pcorpet:
…ption even if persistence failed: there's already create! for that.

[Review on Reviewable](https://reviewable.io/reviews/dynamoid/dynamoid/47)


philipmw:
Thank you.

PR #52, opened 2016-03-02
Expose the truncate method on the adapter, test it and fix it.

pcorpet:
The truncate method was already there, but with no access :-( That's sad because it would be pretty handy for one of my projects.

Although when exposing it I found a bug in it.

[Review on Reviewable](https://reviewable.io/reviews/dynamoid/dynamoid/52)


philipmw:
Thanks. Would you rebase and ship it?

PR #53, opened 2016-03-05
Add config options, use_timestamps and use_namespace

cignoir:
When using Dynamoid for existing tables,  default namespace of tables and fields of timestamps may be annoying. I think those should be selectable as follows.

```ruby
Dynamoid.configure do |config|
  config.use_namespace = false # default: true
  config.use_timestamps = false # default: true
end
```


cignoir:
PR #40 is better about namespacing.

PR #55, opened 2016-03-10
Update dynamoid.gemspec

bglusman:
Am using master locally and building as a gem, and noticed this file is missing :-)


bglusman:
Weird that it failed ALL specs...  what's up with the test failures?  Tests pass for me locally, though 4 pending, and deprecation warning about `last_comment`, though last_comment looked like a failure in the test suite...  let me know if I can help fix!

philipmw:
Thanks!

PR #65, opened 2016-04-12
Fix cache instance variable(@table_name).

takeshinoda:
If there is a cache,  doesn't process the table name .


pboling:

PR #72, opened 2016-05-20
Rake commands to ping DynamoDB and to create tables from the models

NrBT:
I've created a couple of simple rake commands for Rails to: 
1. test the connection to the DynamoDB (local or remote) using the current Dynamoid.configuration settings
2. Create the tables in advance of using them which kinda works around the lag Dynamo has when creating tables

### DIFF
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -13,7 +13,7 @@ module Dynamoid #:nodoc:
       self.read_only_attributes = []
       self.base_class = self
 
-      Dynamoid.included_models << self
+      Dynamoid.included_models << self unless Dynamoid.included_models.include? self
     end
 
     module ClassMethods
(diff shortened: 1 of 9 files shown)


spec/spec_helper.rb
```diff
@@ -4,8 +4,6 @@
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
 $LOAD_PATH.unshift(File.dirname(__FILE__))
 
-MODELS = File.join(File.dirname(__FILE__), "app/models")
```
NrBT: This line is duplicated a little later on

lib/dynamoid/document.rb
```diff
@@ -13,7 +13,7 @@ module Document
       self.read_only_attributes = []
       self.base_class = self
 
-      Dynamoid.included_models << self
+      Dynamoid.included_models << self unless Dynamoid.included_models.include? self
```
NrBT: I saw a case where the same class had been added to the Dynamoid.included_models more than once. I couldn't find a way to reproduce it so haven't written a test proof.

pboling:

pboling:
@NrBT still interested in this?

Issue #74, opened 2016-05-23
Multiple typos in documentation


pboling:
```
      # @example
      #   class User
      #     include Dynamoid::Document
      #     field :email,          :string
      #     field :age,            :integer
      #     field :gender,         :string
      #     field :rank            :number
      #     table :key => :email
      #     global_secondary_index :hash_key => :age, :range_key => :gender
      #   end
      #   User.find_all_by_secondary_index(:age => 5, :range => {"rank.lte" => 10})
```

1. The `:range_key => :gender` should instead be `:range_key => :rank`.
2. Curly braces are required around the first hash param in:
`User.find_all_by_secondary_index(:age => 5, :range => {"rank.lte" => 10})`, as
`User.find_all_by_secondary_index({:age => 5}, :range => {"rank.lte" => 10})`

PR #76, opened 2016-05-24
Fix documentation

pboling:
Fixes https://github.com/Dynamoid/Dynamoid/issues/74

cc @bglusman  - How do I submit PRs to dynamoid-edge?  Where is that fork/repo?

### DIFF
--- a/lib/dynamoid/finders.rb
+++ b/lib/dynamoid/finders.rb
@@ -120,9 +120,11 @@ module Dynamoid
       #     field :gender,         :string
       #     field :rank            :number
       #     table :key => :email
-      #     global_secondary_index :hash_key => :age, :range_key => :gender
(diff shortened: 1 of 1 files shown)


bglusman:
For discoverability, the fork is this repo/branch, and released as needed on dynamoid-edge https://github.com/stellaservice/Dynamoid/tree/dynamoid-edge

Obviously it's versions numbers are already out of sync with Dynamoid, but plan is only to release snapshot updates of merges into master in between official releases.  Could switch to a different versioning scheme after next release, e.g. right now we've released dynamoid-edge 1.1.1 and 1.1.2 ahead of dynamoid, but if they release 1.1.1, perhaps we'll make next master release, if any 1.1.1a and follow each version with letter releases so it's clear what it's relation is.

PR #80, opened 2016-06-21
DynamoDB BatchGet does not work with more than 100 items

lucasas:
https://forums.aws.amazon.com/thread.jspa?threadID=100194

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -94,8 +94,10 @@ module Dynamoid
       #
       # @todo: Provide support for passing options to underlying batch_get_item http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#batch_get_item-instance_method
       def batch_get_item(table_ids, options = {})
-        request_items = Hash.new{|h, k| h[k] = []}
+request_items = Hash.new{|h, k| h[k] = []}
         return request_items if table_ids.all?{|k, v| v.empty?}
+
+        ret = Hash.new([].freeze) # Default for tables where no rows are returned
 
         table_ids.each do |t, ids|
           next if ids.empty?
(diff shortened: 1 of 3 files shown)


lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -38,38 +38,44 @@ def client
       #
       # @todo: Provide support for passing options to underlying batch_get_item http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#batch_get_item-instance_method
       def batch_get_item(table_ids, options = {})
-        request_items = Hash.new{|h, k| h[k] = []}
+request_items = Hash.new{|h, k| h[k] = []}
```
pboling: Please fix spacing.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -38,38 +38,44 @@ def client
...
           rng = tbl.range_key.to_s
 
-          keys = if rng.present?
-            Array(ids).map do |h,r|
-              { hk => h, rng => r }
-            end
-          else
-            Array(ids).map do |id|
-              { hk => id }
+          Array(ids).each_slice(100) do |ids|
```
pboling: Make `100` a configurable option in the `Config` class.  DynamoDB users will all have different paid settings, and some will need to set a smaller batch size.

pboling:

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
peeyush1234: Making it configurable is a good idea. 
@pboling, What are your thoughts on using 100 as default if the batch size config is not provided by the client.

pboling: All I know about batch calls is there is a limit of 25 per batch on creates and deletes.  As this is for get, not create or delete, I am not sure what the default is, or the maximum.  If Amazon has a default batch size, or a recommended batch size we should use that.  If not then 100 seems reasonable.

lucasas:
@pboling can you check if everything is ok?

pboling:
Looks good!

PR #81, opened 2016-07-01
Allow put_item options

alexperto:
Provide support for various options http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#put_item-instance_method


pboling:
Merged as part of #95 and released in 1.2.1

PR #82, opened 2016-07-01
Include batch_write_item API

alexperto:
See:

http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#batch_write_item-instance_method


pboling:
Merged as part of #95 and released in 1.2.1

PR #86, opened 2016-07-08
Fix functionality for associative tables with non-id primary keys

everett-wetchler:


bglusman:
Nice!  I had wanted to use a non-id primary key myself and wound up not partially cause of this issue, thanks for tackling!  The test failures may be flickers, as several ruby versions passed, though do you have any reference for whether this change should always work for all primary keys, id and non-id?  I thought it would be harder to fix than this, but glad it wasn't.  Maybe we can add a spec showing both an id key relation and a non-id primary key relation both working just to document both work.  Thanks for tackling this @everett-wetchler !

everett-wetchler:
Happy to help @bglusman 
The tests in `spec/dynamoid/persisence_spec.rb` do use some models with id primary keys as well as Magazine which I added a non-id primary key to in `app/models/magazine.rb`. Could you help me understand what other test you had in mind?

pboling:

everett-wetchler:
Done

bglusman:
Was about to merge this in but realized I never responded to @everett-wetchler's question, I think if there are tests covering both use cases that's sufficient, though I'm tempted to bring them together in a single file to make clear they're both supported...  what do you think @pboling, merge as is or any better spec/documentation to add for this?

pboling:
I am ambivalent about if they are in the same file, though it would be nice.  I don't really have an opinion on splitting specs into many files per use case and combining related scenarios.

everett-wetchler:
Unless there's a strong opinion I'd prefer to go ahead and merge as is.

PR #92, opened 2016-08-01
Support to Rails 5

noelrocha:
* Changed gemspec to support rails 5.
* Added **active_model_serializers** into gemspec
* Removed **include ActiveModel::Serializers::Xml** from *Components.rb* because it wont work on rails 5

### DIFF
--- a/lib/dynamoid/components.rb
+++ b/lib/dynamoid/components.rb
@@ -23,7 +23,7 @@ module Dynamoid
     include ActiveModel::Naming
     include ActiveModel::Observing if defined?(ActiveModel::Observing)
     include ActiveModel::Serializers::JSON
-    include ActiveModel::Serializers::Xml
+    #include ActiveModel::Serializers::Xml
     include Dynamoid::Fields
     include Dynamoid::Indexes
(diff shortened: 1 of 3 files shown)


bglusman:
I think this looks good, but a) please delete the XML serialzier rather than comment out, and b) Can you look into the test failure, see if it reproduces locally etc and try and fix?  Thanks!

pboling:

lib/dynamoid/components.rb
```diff
@@ -23,7 +23,7 @@ module Components
     include ActiveModel::Naming
     include ActiveModel::Observing if defined?(ActiveModel::Observing)
     include ActiveModel::Serializers::JSON
-    include ActiveModel::Serializers::Xml
+    #include ActiveModel::Serializers::Xml
```
bglusman: I understand this was removed from Rails 5, but commenting it out is not appropriate.  probably ` include ActiveModel:: Serializers if defined?(ActiveModel:: Serializers)` is the easiest fix here, as done above for ActiveModel::Observing

PR #95, opened 2016-08-07
Merge pboling/master to Dynamoid/master

pboling:
pboling/master has been running intricately.com's data backend for months, and has some enhanced features, like batch writes, and batch deletes.  Thanks to @alexperto for much of the work.


bglusman:
@pboling oh hadn't seen this when I commented in other issue but I guess this has same test issue?  We could try a CI besides Travis and see if we can get a more consistent build elsewhere, but still looking over the diff, you've been busy!

Gemfile.lock
```diff
@@ -0,0 +1,91 @@
+PATH
```
bglusman: Don't think we should be checking in a Gemfile.lock to a gem...  unless guidance on that has changed or you have a specific reason it's needed?  Could have locally or as an example to share among devs maybe if any specific concerns, but I'm citing Yehuda [here](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) for reference

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -463,7 +494,7 @@ def expected_stanza(conditions = nil)
         expected = Hash.new { |h,k| h[k] = {} }
         return expected unless conditions
 
-        conditions[:unless_exists].try(:each) do |col|
+        conditions.delete(:unless_exists).try(:each) do |col|
```
bglusman: Unsure if some reason why not at a glance, but should the `condiitions[:if]` below also be a `delete` for symmetry/POLA purposes perhaps?

Gemfile.lock
pboling: I agree.  This needs to be removed and gitignored.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
pboling: Yes, and that it is not is actually probably causing half those spec failures...  Thanks for noticing!

bglusman:
Also just noticed @alexperto has the batch_write_item PR open, did you pull this from his branch? don't want him to lose credit for that so if you need to make any changes to his implementation maybe you can base off his branch so he still has his commits there, same for anyone else's work that's in this branch.

PR #101, opened 2016-09-06
Rails 5 support

jihaia:
Minimum updates needed to allow Rails 5.0.0.x the ability to include Dynamoid.

### DIFF
--- a/lib/dynamoid/components.rb
+++ b/lib/dynamoid/components.rb
@@ -23,7 +23,6 @@ module Dynamoid
     include ActiveModel::Naming
     include ActiveModel::Observing if defined?(ActiveModel::Observing)
     include ActiveModel::Serializers::JSON
-    include ActiveModel::Serializers::Xml
     include Dynamoid::Fields
     include Dynamoid::Indexes
     include Dynamoid::Persistence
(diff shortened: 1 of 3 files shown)


tadast:
FYI I've tried this set of changes and it seems to work well with Rails 5. Not sure why the tests are failing on travis though, seems like a setup issue?

jihaia:
The failing tests occur on master - or at least on the fork I took. Happy to help review and track down the issues, but they appear unrelated to the proposed changes.

pboling:
Tests on master are fixed.  Tests are failing here because you have updated the gems to version that do not work on older versions of Ruby.  Will need to utilize a tool like `appraisals` to make this work across the supported ruby and rails versions.

pboling:
Please rebase on latest master, and fix the spec suite by implementing appraisals.

lib/dynamoid/components.rb
```diff
@@ -23,7 +23,6 @@ module Components
     include ActiveModel::Naming
     include ActiveModel::Observing if defined?(ActiveModel::Observing)
     include ActiveModel::Serializers::JSON
-    include ActiveModel::Serializers::Xml
```
bglusman: see comments on the other 2 rails 5 fixup PR's :-)

PR #102, opened 2016-09-07
Fix AWS SDK V2 query to pull in all results

richardhsu:
- [Fix] Previously query would only return the first page of data which
  is a maximum of 1MB by DynamoDB. This fix loops to retrieve all
  results beyond the 1MB limit

Note on styling: Rewrote to be similar as `scan` defined below.


pboling:
Thanks!

PR #103, opened 2016-09-08
Add a not about test setup

tadast:
Add an example of database cleaning strategy for test environment ([inspired by this](https://github.com/Dynamoid/Dynamoid/blob/a1141e7f832078c3c1b2e1823d2f251e99e237b4/spec/dynamodb_local.rb#L40-L48)) and a note about setting a different namespace in test init. 

The latter is important to prevent development data deletion when `RAILS_ENV` is not set to 'test' for any reason.


pboling:
Thanks!

PR #64, opened 2016-04-12
Raw datatype

psergus:
we really needed native type that picks up existing models on DynamoDB. AWS's SDK V2 already has a support for persisting datatypes to dynamo, so why not utilize it.

PR #104, opened 2016-09-13
Raw datatype

psergus:
we really needed native type that picks up existing models on DynamoDB. AWS's SDK V2 already has a support for persisting datatypes to dynamo, so why not utilize it.
This is rebased PR in favor of closed #64


pboling:
Thank you!

pboling:
@dahnny012 this hasn't been released yet.  I will create a new release will all the latest work in master soon.

PR #69, opened 2016-05-03
Rails 5 Serialization issue resolved

josephverbeck:
Fixed issue with serialization in rails 5.  The problem was serialization had been moved to it's own gem which required adding it to the spec gem.  It also needed to be required in the components.rb file.  Hopefully this helps you as rails 5 is released!

PR #126, opened 2017-01-17
Rails 5 compatibility, and empty tables default

rall:
This replicates [a stale pull request](https://github.com/Dynamoid/Dynamoid/pull/69) against the Dynamoid project. We need this functionality now, so we should use our own fork until the official repo is caught up.

Makes the gem rails 5 compatible, and adds a sensible default value from the `tables` method when the dynamodb tables haven't yet been created


rall:
sorry! wrong repo!

PR #140, opened 2017-04-07
default value should be per instance

thilo:
was using the same references for all instances before as it was initialized on the class.


pboling:
looks good

PR #143, opened 2017-04-11
Prevent load multiple times

sselvamani22:
When I start server, MODEL constant value initializing multiple times.


pboling:
Looks good!

PR #146, opened 2017-06-07
Update Readme.md - add description of `raw` type

andrykonchin:
Readme missing information about `raw` type, so I have added `raw` into list of correct types and added following short description based on specs:

> `raw` type means you can store Ruby Array, Hash, String and numbers.


pboling:
Thank you!

PR #147, opened 2017-06-07
Fix warning about scan operation

andrykonchin:
When model makes scan and  `Dynamoid::Config.warn_on_scan option` is true (by default) we log some warnings.

We try to log a list of attributes to build index on, but actually we log all the model field definitions:

>You can index this query by adding this to adapter.rb: index [:[:created_at, {:type=>:datetime}], ...

Assume we have to log attributes from a query passed to `where` method.

e.g. for the code
`User.where(name: 'x', password: 'password').all`

We will log
`# => ... user.rb: index [:name, :password]`

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -139,7 +139,7 @@ module Dynamoid #:nodoc:
       def records_via_scan
         if Dynamoid::Config.warn_on_scan
           Dynamoid.logger.warn 'Queries without an index are forced to use scan and are generally much slower than indexed queries!'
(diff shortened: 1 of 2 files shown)


pboling:
I actually don't know if any of that is still relevant.  I don't use the ActiveRecord features of Dynamoid at all (strange, I know, and really not the best for the only active maintainer of the gem).

I do recall something about those types of indexes having their support changed (dropped?) by Amazon.  So the warning, and the feature itself, may be a complete red herring.

Need someone who is using that part of the tool to do some digging.

pboling:
@andrykonchin did you see my previous comment? ^

pboling:
Going to merge this, because it is an improvement.  Still I am not sure if the warning if correct at all.

PR #148, opened 2017-06-07
Add Chain#last method

andrykonchin:
We have `Chain#first` method but don't have `Chain#last`.

Now we can do
`User.where(admin: true).last`

It may be not useful in production code but looks good in specs


pboling:
Thanks

PR #150, opened 2017-06-08
Support field accessors overriding by extentions

andrykonchin:
Currently when you declare fields in a model we generate fields accessors right in the model class.
If we want to write an extension (as a separate module) which should override field getter/setter and add some functionality we aren't able to do this.

Example of such extension is `simple_enum` project (https://github.com/lwe/simple_enum).

There is a solution. We can find it in the `mongoid` project: getters and setters declare in a separate module which includes as early as possible so any other module is able to override generated methods.

`mongoid` approach:
module creation - https://github.com/mongodb/mongoid/blob/master/lib/mongoid/fields.rb#L412
dynamic method declaration - https://github.com/mongodb/mongoid/blob/master/lib/mongoid/fields.rb#L412 

Example of using extension:

```ruby
module Extention
  def name
    super.upcase
  end

  def name=(s)
    super(s.try(:downcase))
  end
end

class User
  include Dynamoid::Document
  include extention

  field :name
end

user = User.new(name: 'AB cd')
user.name
#=> 'AB CD'

```

### DIFF
--- a/lib/dynamoid/associations.rb
+++ b/lib/dynamoid/associations.rb
@@ -19,7 +19,7 @@ module Dynamoid
     
     # Create the association tracking attribute and initialize it to an empty hash.
     included do
-      class_attribute :associations
+      class_attribute :associations, instance_accessor: false
       
       self.associations = {}
     end
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -8,7 +8,7 @@ module Dynamoid #:nodoc:
     include Dynamoid::Components
 
     included do
-      class_attribute :options, :read_only_attributes, :base_class
+      class_attribute :options, :read_only_attributes, :base_class, instance_accessor: false
       self.options = {}
       self.read_only_attributes = []
       self.base_class = self
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -120,6 +120,10 @@ module Dynamoid #:nodoc:
     #
     # @since 0.2.0
     def initialize(attrs = {})
+      # we need this hack for Rails 4.0 only
+      # because `run_callbacks` calls `attributes` getter while it is still nil
+      @attributes = {}
+
       run_callbacks :initialize do
         @new_record = true
         @attributes ||= {}
--- a/lib/dynamoid/fields.rb
+++ b/lib/dynamoid/fields.rb
@@ -15,7 +15,7 @@ module Dynamoid #:nodoc:
 
     # Initialize the attributes we know the class has, in addition to our magic attributes: id, created_at, and updated_at.
     included do
-      class_attribute :attributes
+      class_attribute :attributes, instance_accessor: false
       class_attribute :range_key
 
       self.attributes = {}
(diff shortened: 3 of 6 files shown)


pboling:
This is really good!  Will merge once the build is passing!

pboling:
@andrykonchin 
Unfortunately the *only* project I use this on is a Rails 4.0 project, and I am the only active maintainer of this project.  Can you explain more about this breaking change?  Are there upgrade instructions, or will this make the gem entirely incompatible with Rails 4.0?

andrykonchin:
@pboling Maybe my last commit description is not clear enough.

I meant the commit "Fix flaky specs." fixes one issue but introduce another one with Rails 4.0. Next commit "Fix document initializing" fixes support of Rails 4.0.

All tests pass with Rails 4.0 so it should work well.

pboling:
@andrykonchin great!  Thanks!  Are you interested in becoming a maintainer?

andrykonchin:
@pboling hmm, not now. At least I need to learn more about `DynamoDb` and dig dipper into `Dynamoid`'s source code. I have no much experience with `DynamoDb`.

Now we transfer our project from mongo to dynamo. If we have a good experience and I become more familiar with with it... Yep, it's possible to become a maintainer.

rromanchuk:
@andrykonchin @pboling  is overriding field setters still safe thing to do? This is kind of sloppy, but i want to override a setter to detect an AR record,  as a convenience that builds a simple hash partial from the model. 

```ruby
module Partialable
  def user=(u)
    u.is_a?(User) ? super({id: u.id, name: u.name}) : super(u)
  end
end


class Post
  include Dynamoid::Document
  include Partialable

  field :user, :map
end
```

```ruby
ar_user = User.last
post = Post.new
# Method user= generated for the field user overrides already existing method
post.user = ar_user
```

pboling:
@andrykonchin actually did end up taking over most of the maintenance of this gem ♥️

I wonder what he thinks about doing this still!

andrykonchin:
@rromanchuk 

Yeah, overriding setters/getters in included module (`Partialable` in the example above) is completely safe. This feature was added [here](https://github.com/Dynamoid/dynamoid/pull/150) and the added specs might be useful.

Dynamoid imitates API of Rails' model so instead of calling `super` it's possible to use `write_attribute` and its alias `[]=` ([documentation](https://rubydoc.info/gems/dynamoid/3.6.0/Dynamoid/Fields#write_attribute-instance_method))

rromanchuk:
@ananthakumaran cool thx

Random question, it looks like `created_at` is ignored when explicitly passed as an attribute to create. This might be a feature not a bug and I'd like to keep those columns managed by dynamoid if possible. The reason i need to rewrite history in this very narrow context is because i'm backfilling/migrating data to dynamoid managed tables. Any thought on what the best way to go about this?

rromanchuk:
actually i'm not sure if what i'm saying is correct.  just realized i was coercing with `to_i` when that's something `:datetime` will do for me

andrykonchin:
Hmm, it should work well and it's easy to check:

```ruby
        created_at = updated_at = Time.now + 1.hour

        obj = klass.create(created_at: created_at, updated_at: updated_at)

        expect(obj.created_at.to_i).to eql(created_at.to_i)
        expect(obj.updated_at.to_i).to eql(updated_at.to_i)
```

Issue #160, opened 2017-06-29
delete_table and create_table in Sync mode not waiting for action to finish (not really sync).


mirokuxy:
as the title says.

PR #161, opened 2017-06-29
fixes the issue of delete-table and create-table not waiting until ac…

mirokuxy:
…tion finish in Sync mode;


mirokuxy:
related issue: #160

PR #162, opened 2017-07-03
Added Rails.root. Unable to find models without it.

gaiottino:
The rake task would not work for me until I added `Rails.root` to the `File.join`. This was true on two different machines running Rails 5.

Not sure what else you want to know for this minor change :)

### DIFF
--- a/lib/dynamoid/tasks/database.rake
+++ b/lib/dynamoid/tasks/database.rake
@@ -5,7 +5,7 @@ namespace :dynamoid do
   desc "Creates DynamoDB tables, one for each of your Dynamoid models - does not modify pre-existing tables"
   task :create_tables => :environment do
     # Load models so Dynamoid will be able to discover tables expected.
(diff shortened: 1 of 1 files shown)


andrykonchin:
There is a config option `models_dir`:

`option :models_dir, :default => "app/models"`

As far as I know `Dynamoid` can be used in non-`Rails` environment. There are a lot of checks if `Rails` constant defined.

So I believe the correct solution is changing of default value - current value for non-`Rails` environment and `./app/models` for `Rails` one.

I am not sure it should be fixed at all. Maybe adding description of `models_dir` option is enough.

pboling:
It seems that this is equivalent to doing `Dynamoid::Config.models_dir = File.join(Rails.root, ...)` and I think that is the intent of the configuration feature.  It allows people to specify the directory to use for models to be wherever they want it to be.  We should not assume a default base directory.

So if there is a bug it is that the default value of `models_dir` needs to include `Rails.root`, not that we should always prepend it.  The settings should work with absolute paths.

Issue #6, opened 2015-06-24
Range finding functionality ('field.gt' et al) works only for Query, not Scan


philipmw:
AWS SDK v2 supports an optional `scan_filter` on Scan API calls, with the same semantics as `query_filter` on Query API calls.

Dynamoid offers convenience methods for finding on range fields, such as `my_chain.where("rangefield.gt" => 5)`, but it works only on Query API calls. On Scan, Dynamoid _allows_ this query chain, but it does not translate the ".gt" component, so you end up scanning for a non-existent field.

We need the range finding functionality to be consistent between Query and Scan.

afgarcia86:
Any update on this?

pboling:
@afgarcia86 this is supported in the latest master.  A new release will be out soon.

PR #167, opened 2017-07-12
Improve querying with scan and query

andrykonchin:
Improving:
- use Query instead of Scan if there are no conditions for sort (range) key in `where` clause
- allow all (provided by DynamoDB) condition operators for non-key fields for Query request (added `in`, `contains`, `not_contains`)
- allow all (provided by DynamoDB) condition operators for Scan request - was available only equality condition

So now we can do following queries:

```ruby
Model.where(partition_key: '', 'field1.in': [1, 2, 3]) # => use Query
Model.where('field1.gt': 10, 'field2.begins_with': 'abc') # => use Scan
```

I continue to use deprecated KeyConditions, QueryFilter and ScanFilter. They are limited and should be replaced with KeyConditionExpression and FilterExpression.

Related issue https://github.com/Dynamoid/Dynamoid/issues/6

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,28 @@ Address.all(batch_size: 100).each { |address| address.do_some_work; sleep(0.01) 
 Address.eval_limit(10_000).batch(100). each { … } #batch specified as part of a chain
 ```
 
+You are able to optimize query with condition for sort key. Following
+operators are available: `gt`, `lt`, `gte`, `lte`, `begins_with`, `between` as well as equality:
+
+```ruby
+Address.where(latitude: 10212)
+Address.where('latitude.gt': 10212)
+Address.where('latitude.lt': 10212)
+Address.where('latitude.gte': 10212)
+Address.where('latitude.lte': 10212)
+Address.where('city.begins_with': 'Lon')
+Address.where('latitude.between': [10212, 20000])
+```
+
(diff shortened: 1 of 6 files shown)


andrykonchin:
The last commit fixes failed spec for `database_timezone` config option. It should fix builds of the all other open PRs.

The last build is failed because of some issues in TravisCI.

andrykonchin:
@pboling any chance to get your feedback?

pboling:
Thanks!  I was on my honeymoon, just got back to my computer.  This looks great!!

andrykonchin:
@pboling 👍 

I don't work with indices but I noticed there are several methods like `find_all_by_secondary_index`.
We cannot use indices with `where` clause and have two separate interfaces - `where` and finders.

Is it a good idea to allow to specify index (global/secondary) for `where` clause? I could be like this:

`Model.with_index(:MyCoolIndex).where(condition)`

pboling:
Oh, that is very interesting.  Unfortunately `find_all_by_secondary_index` is literally the only part of Dynamoid I use.  I don't use associations, or Railsy parts at all, so I am not sure how much people might want it.  It *sounds* like it would be useful to me.

andrykonchin:
@pboling https://github.com/Dynamoid/Dynamoid/issues/6 can be resolved

Issue #138, opened 2017-03-28
Range Finding example fix


archonic:
The readme has this as an example:

    User.where("created_at.gt" => DateTime.now - 1.day).all

This caused `ArgumentError: unsupported type, expected Hash, Array, Set, String, Numeric, IO, true, false, or nil, got DateTime` for me. I had to convert to integer instead.

    User.where("created_at.gt" => 1.day.ago.to_i).all

archonic:
Upon inspection, it seems dynamoid can't find anything involving a date range query. Does the date range example work for anyone? I'll have to use the AWS SDK directly if not. Kinda defeats the purpose...

pboling:
That may be due to changes I made in the latest release from a few days ago.  I'll take a look and see if I broke something.

mirokuxy:
still not fixed?

pboling:
There has been no PR to fix this.  I haven't had time to look into it.  I'll see if I can figure something out.

richardhsu:
Can close this issue :) Thanks @andrykonchin, needed this too!

pboling:
@mirokuxy fixed!

Issue #46, opened 2016-02-24
Range finding doesn't work as documented


ericpeters0n:
For

```
class MyObject
  include Dynamoid::Document
  field :name
end
```

this:

```
MyObject.where("created_at.gt" => DateTime.now - 1.day).all
```

yields

```
ArgumentError: unsupported type, expected Hash, Array, Set, String, Numeric, IO, true, false, or nil, got DateTime
```

Am I missing something?

ericpeters0n:
@sumocoder  - Appreciate your work on the type mappings... Any hints here?

ericpeters0n:
@philipmw - It looks like @sumocoder and @ayemos have PRs that address this breakage to some degree...

ericpeters0n:
Ah, appears this is a result of #17

afgarcia86:
I have the same issue...

richardhsu:
This works now based on @andrykonchin's PR, I've been testing this in a staging environment and it works well.

PR #170, opened 2017-07-26
Add type casting in `where` clause

andrykonchin:
Cast values in `where` clause to declared field types

Fix open issues https://github.com/Dynamoid/Dynamoid/issues/46 and https://github.com/Dynamoid/Dynamoid/issues/138

---

Looks like this PR breaks existed behaviour - search for field with custom type by `casted` value:

```
field :price, MoneyInstanceDump

Model.where(price: '5.0') 
```

It worked because we store value `5` in the storage and bypass type casting.
Now it shouldn't work because type casting requires something like this:

```
Model.where(price: MoneyInstanceDump.new('5.0')) 
```

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -145,7 +145,7 @@ module Dynamoid #:nodoc:
       def records_via_scan
         if Dynamoid::Config.warn_on_scan
           Dynamoid.logger.warn 'Queries without an index are forced to use scan and are generally much slower than indexed queries!'
-          Dynamoid.logger.warn "You can index this query by adding this to #{source.to_s.downcase}.rb: index [#{query.keys.sort.collect{|attr| ":#{attr}"}.join(', ')}]"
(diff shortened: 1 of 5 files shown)


lib/dynamoid/criteria/chain.rb
```diff
@@ -175,8 +176,8 @@ def range_hash(key)
       end
 
       def field_hash(key)
-        val = query[key]
         attr, operation = key.to_s.split('.')
```
pboling: We should change it here as well.

lib/dynamoid/criteria/chain.rb
```diff
@@ -156,9 +156,10 @@ def records_via_scan
       end
 
       def range_hash(key)
-        val = query[key]
+        attr, operation = key.to_s.split('.')
```
andrykonchin: @pboling please show me this in the Rails documentation. Haven't found it.

pboling: @andrykonchin Sorry, I misread my documentation tool.  It is core Ruby, not Rails.  It is defined in the core Ruby `Module` object.

http://ruby-doc.org/core-2.4.1/Module.html#method-i-attr

andrykonchin: 👍

andrykonchin:
Done

Issue #60, opened 2016-03-24
Invalid item size limit calculations and MAX_ITEM_SIZE value


activefx:
Dynamoid sets the `MAX_ITEM_SIZE` value to 64 KB. However, the current limit is 400 KB. Please see the [Limits in DynamoDB](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-items) documentation. 

Also, the `#write_attribute` method within `Fields` incorrectly calculates this limit. 

``` ruby
def write_attribute(name, value)
  if (size = value.to_s.size) > MAX_ITEM_SIZE
    Dynamoid.logger.warn "DynamoDB can't store items larger than #{MAX_ITEM_SIZE} and the #{name}  field has a length of #{size}."
  end
...
```

The calculation for the warning only takes into account the size of a single value. However the item limit is based upon the sum of all key value pairs within the item. If you wanted to keep the warning, to be accurate, the calculation would have to be something like: 

``` ruby
attributes.map{|k,v| "#{k}#{v}"}.join.size
```

I can submit a pull request if you let me know the direction you would like to take.

pboling:
Pull requests, and advice are appreciated.  This project is now under new maintainers.  What do *you* think would be the best way to fix this?

andrykonchin:
Done

PR #171, opened 2017-07-26
Remove unuseful warning about item size exceeding limits

andrykonchin:
Dynamodb has some limitations for item and fields size but there is no rules to calculate item size in official documentation

I have found only this explanation:
https://stackoverflow.com/a/9018008/219594

So I propose completely remove any logic related to Dynamoid's size limitations like other ORMs don't do this for RDMSs

Open issue https://github.com/Dynamoid/Dynamoid/issues/60

### DIFF
--- a/lib/dynamoid.rb
+++ b/lib/dynamoid.rb
@@ -36,8 +36,6 @@ end
 module Dynamoid
   extend self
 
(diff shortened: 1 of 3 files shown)


pboling:
Agreed.

Issue #85, opened 2016-07-07
eval_limit and batch don't do what I expect


JohnSmall:
I'm not sure what the chain eval_limit and batch should do e.g. 

```
`SomeModel.eval_limit(5).batch(5)`
```

Should that get 5 records, in a batch of 5, or 5 batches of 5 records each?

When I tried it it scanned the entire table ignoring the limit.

pboling:
@andrykonchin do you have any idea about this?

andrykonchin:
@pboling 

`Model.where({}).eval_limit(n)` means we load first `n` items

`Model.where({}).batch(m)` means we load all matched items but internally paginate them and load `m` items at a time.

`Model.where({}).eval_limit(n).batch(m)` chain completely ignores limit `n` and works like `batch` call. But yes, I was expected behavior described by @JohnSmall too.

pboling:
So, it is a bug 👍 , We'll leave this open to track it until we fix it.

richardhsu:
This is updated again in https://github.com/Dynamoid/Dynamoid/pull/183 to handle scan limit as well.

pboling:
@richardhsu so #183 doesn't fix this bug, but it adds another type of limit "support" that will be ignored if `batch` is chained?

```
Model.where({}).eval_limit(n).batch(m)
Model.where({}).scan_limit(n).batch(m)
```
Both limit types are ignored?

richardhsu:
@pboling can you give me examples that you're seeing don't work?

I'm seeing it respect the `scan_limit` and `record_limit` and making batch queries to DynamoDB with `batch_size` assuming `batch_size < scan_limit` or `batch_size < record_limit`.

```
# Record Limit with batch of 3
> model.where({}).record_limit(5).batch(3).to_a.count
REQUEST {:table_name=>"dynamoid_tests_customer", :limit=>3}
REQUEST {:table_name=>"dynamoid_tests_customer", :limit=>2, :exclusive_start_key=>{"age"=>#<BigDecimal:3e64828,'0.3E1',9(18)>, "id"=>"1"}}
=> 5

# Record limit with batch of 1
> model.where({}).record_limit(5).batch(1).to_a.count
REQUEST {:table_name=>"dynamoid_tests_customer", :limit=>1}
[...]

PR #172, opened 2017-08-09
AWS SDK: Update Query/Scan to obey limits

richardhsu:
- [Update] Query did not obey limit because it would keep
  looping as long as there were pages to read. Added count
  to return once limit was reached based on scanned records.
- [Update] Scan did not obey limit when in batching mode
  and would have similar issue as Query. Added count to return
  once limit was reached based on scanned records.

Open issue https://github.com/Dynamoid/Dynamoid/issues/85

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -494,9 +494,16 @@ module Dynamoid
         q[:query_filter]   = query_filter
 
         Enumerator.new { |y|
+          eval_count = 0
           loop do
             results = client.query(q)
             results.items.each { |row| y << result_item_to_hash(row) }
+
+            # If limit is set then keep track and stop requesting once
+            # completed. This is eval_limit so we compare to scanned_count
+            # rather than resulting data
+            eval_count += results.scanned_count
+            break if q[:limit] && eval_count >= q[:limit]
 
             if(lk = results.last_evaluated_key)
               q[:exclusive_start_key] = lk
@@ -539,12 +546,20 @@ module Dynamoid
 
         Enumerator.new do |y|
           # Batch loop, pulls multiple requests until done using the start_key
+          # by paging in by batch size rather than limit size
+          eval_count = 0
           loop do
             results = client.scan(request)
-
             results.data[:items].each { |row| y << result_item_to_hash(row) }
 
-            if((lk = results[:last_evaluated_key]) && batch)
+            # If limit is set then keep track and stop requesting once
+            # completed. This is eval_limit so we compare to scanned_count
+            # rather than resulting data
+            eval_count += results.scanned_count
+            break if limit && eval_count >= limit
+
+            # Keep pulling if we haven't finished paging in all data
+            if(lk = results[:last_evaluated_key])
               request[:exclusive_start_key] = lk
             else
               break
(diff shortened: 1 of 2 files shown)


pboling:
Great!

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -539,12 +546,20 @@ def scan(table_name, scan_hash, select_opts = {})
...
           loop do
             results = client.scan(request)
-
             results.data[:items].each { |row| y << result_item_to_hash(row) }
 
-            if((lk = results[:last_evaluated_key]) && batch)
+            # If limit is set then keep track and stop requesting once
+            # completed. This is eval_limit so we compare to scanned_count
+            # rather than resulting data
+            eval_count += results.scanned_count
```
andrykonchin: Looks like `scanned_count` returns count of loaded items before filtering. So if we provide some conditions with `where` clause `eval_limit` may return items less that was requested.

Should we use actual data size here?

richardhsu: I was not sure about this, considering the code was changed from 'limit' to 'eval_limit' I assumed this change was meant to limit number of records DynamoDB scanned (similar to DynamoDB's definition of limit) rather than actual limit of the data returned. I can see it being fairly bad to do regular 'limit' if we end up having a scan since potentially scanning over all data. Which is why I implemented against 'scanned_count'.

On the flip side, I can see how if filters remove items then if you get an empty set, you are a bit stuck with not being able to query more, I think the TODO in that case would mean facilitating a way to return the Last Evaluated Key so could actually continue paginating through results. But not sure best way to do this with the general style that these queries may be written in. Maybe a class variable would work best to store it with thread safety considerations.

andrykonchin: As far as `limit` doesn't work for a long time we can consider it as a new feature.

For me it would be a surprise to set `limit` and receive fewer collection. But for performance reasons as you said user may need to limit scanned items count.

Maybe we can support both types of limits? Result limit and scan limit?

And ability to set `offset` primary key (aka `ExclusiveStartKey`)

richardhsu: That sounds good, so why don't we have:

* `eval_limit` - specifies the limit of returned records
* `scan_limit` - specifies the limit of scanned records (new)

In many cases these will act the same but in cases of filtering, `eval_limit` will continue until there are no more records or `eval_limit` is hit whereas the `scan_limit` will limit to basically one query (or multiple if used with `batch`).

Finally if `eval_limit` and `scan_limit` are used together, then we have following cases:

* `eval_limit = scan_limit` then our query `limit` will be the same
* `eval_limit > scan_limit` then our query `limit` will be the `scan_limit` (takes minimum)
* `eval_limit < scan_limit` then our query `limit` will be the `eval_limit` (takes minimum) but we stop once `scan_limit` is reached or when `eval_limit` is reached, whichever comes first.

If this sounds good, I can put this PR together to update to these semantics and add to README.

As for the `offset`, we already have that with `start` function, the only issue is things are okay if your result set includes your scanned set. However, if your scanned set is the super set then we need to return the `LastEvaluatedKey` somehow. It's not currently exposed by the `aws_sdk_v2` query/scan methods but dealing with this can be a separate issue.

andrykonchin: Sounds good :+1: 
With Rails background name `eval_limit` looks strange for me. Maybe we need an alias `limit` for it.

@pboling What do you think about adding `scan_limit`?

pboling: @andrykonchin @richardhsu 
I like the idea of splitting the limits and having separate names.  `scan_limit` is very clear for that concept.

I like the idea of *not* using the familiar `limit` because then people won't assume an incorrect meaning if they don't read the documentation.  Since you pay for what you use this matters.  People could assume `limit` will prevent reading more than `limit` records, or that it will read as much as needed until it finds `limit` records.

As such I like using something differentiated from the standard `limit` token.  I agree that `eval_limit` is strange, because of the *eval is evil* mantra, and there is no Ruby eval happening here.

Other potential ideas:
• `result_limit`
• `record_limit`

PR #173, opened 2017-08-09
Chain: Allow querying Global/Local Secondary Index

richardhsu:
- [Update] Allow in chaining to query and underlying implementation
  utilizes global/local secondary indices that may be available
  instead of doing a scan. Thus we can query subsets of data instead
  of using the full `find_all_by_secondary_index`.
- [Update] Added serialized types to be able to be range_key and
  types for dump for query since serialized is just a string.
  Caveat to developer would be they need to know would be range
  type comparisons would operate on string comparisons! So if
  using serialized form then understand sort is lexigraphically
  sorted!

Main reason is I want to be able to query entities but *do not* want to enumerate all records at once. Thus I need to use `eval_limit` and building it in chaining works out nicely with parallels to ActiveRecord.

This would then replace

```
find_all_by_secondary_index(
    {
        dynamo_primary_key_column_name => dynamo_primary_key_value
    }, # The signature of find_all_by_secondary_index is ugly, so must be an explicit hash here
    :range => {
        "#{range_column}.#{range_modifier}" => range_value
    },
    # false is the same as DESC in SQL (newest timestamp first)
    # true is the same as ASC in SQL (oldest timestamp first)
    :scan_index_forward => false # or true
)
```

to become the following

```
where(:dynamo_primary_key_column_name => dynamo_primary_key_value,  "#{range_column}.#{range_modifier}" => range_value).all
```

but now we can take advantage of chaining instead of having to build in all the extra opts into `find_all_by_secondary_index` which also would kind of be counter-intuitive to be doing a limit in.

If this looks good we can also replace the implementation of the finders to use the chain implementations (But figured open this up for feedback). Maybe can use with more explicit checks on top for `MissingIndex` errors since the chain implementation does best effort which is similar to how SQL would have performed things with or without indices.

Thanks!

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -456,6 +456,10 @@ module Dynamoid
             hk  => token[:hash_key_element],
             rng => token[:range_key_element]
           }
+          # For secondary indices the start key must contain the indices composite key
+          # but also the table's composite keys
+          q[:exclusive_start_key][table.hash_key] = token[:table_hash_key_element] if token[:table_hash_key_element]
+          q[:exclusive_start_key][table.range_key] = token[:table_range_key_element] if token[:table_range_key_element]
         end
 
         key_conditions = {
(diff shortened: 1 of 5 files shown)


richardhsu:
Oops, must have broke something upon squashing my commits for this, will look into this and fix tomorrow.

pboling:
This looks really good!  I will try to test it out over the next few days on a large DynamoDB instance and make sure the query results are good.

PR #174, opened 2017-08-10
allow batch_get_item for integer partition key

mudasirraza:
If partition key is of type number, it throws NoMethodError: undefined method `empty?' for 1:Fixnum. This PR fixes this scenario.


pboling:
Needs a spec I think, but mill merge now as it looks like a very simple fix.

PR #179, opened 2017-08-21
Store dates as days count since 1970-01-01

andrykonchin:
Currently we have an issue with storing `date` fields.

`date` fields are stored as Unix timestamps. We convert date to time and vise versa so take into account current time zone. If system time zone changes we can get shift +/- 1 day

I propose to store not seconds but days since Unix epoch `01-01-1970`

### DIFF
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -13,6 +13,8 @@ module Dynamoid
     attr_accessor :new_record
     alias :new_record? :new_record
 
+    UNIX_EPOCH_DATE = Date.new(1970, 1, 1)
+
     module ClassMethods
 
       def table_name
@@ -135,7 +137,7 @@ module Dynamoid
                 if value.is_a?(Date) || value.is_a?(DateTime) || value.is_a?(Time)
                   value.to_date
                 else
-                  Date.new(1970, 1, 1) + value.to_i
+                  UNIX_EPOCH_DATE + value.to_i
                 end
               when :boolean
                 # persisted as 't', but because undump is called during initialize it can come in as true
@@ -177,7 +179,7 @@ module Dynamoid
             when :datetime
               !value.nil? ? value.to_time.to_f : nil
             when :date
-              !value.nil? ? (value.to_date - Date.new(1970, 1, 1)).to_i : nil
+              !value.nil? ? (value.to_date - UNIX_EPOCH_DATE).to_i : nil
             when :serialized
               options[:serializer] ? options[:serializer].dump(value) : value.to_yaml
             when :raw
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -13,7 +13,7 @@ module Dynamoid
     attr_accessor :new_record
     alias :new_record? :new_record
 
-    UNIX_EPOCH_DATE = Date.new(1970, 1, 1)
+    UNIX_EPOCH_DATE = Date.new(1970, 1, 1).freeze
 
     module ClassMethods
 
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -135,7 +135,7 @@ module Dynamoid
                 if value.is_a?(Date) || value.is_a?(DateTime) || value.is_a?(Time)
                   value.to_date
                 else
-                  Time.at(value).to_date
+                  Date.new(1970, 1, 1) + value.to_i
                 end
               when :boolean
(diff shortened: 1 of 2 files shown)


richardhsu:
So does this occur if you configure Dynamoid to use a different time zone for the dump data?

welkie:
In the [appendix of the official documentation](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/AppendixSampleTables.html), they have an example table including data types and indexes. They use strings to store dates and datetimes, storing them as an ISO 8601 date locked to UTC:

```json
{
    "PutRequest": {
        "Item": {
            "Id": {
                "S": "Amazon DynamoDB#DynamoDB Thread 1"
            },
            "ReplyDateTime": {
                "S": "2015-09-15T19:58:22.947Z"
            },
            "Message": {
                "S": "DynamoDB Thread 1 Reply 1 text"
            },
            "PostedBy": {
                "S": "User A"
            }
        }
    }
}
```

I really like this approach to storing dates when the database doesn't have a specific type for them. It's objective and strings support all the query options you would need (like a range query). I think it would make sense to parse them into date types in Dynamoid and then have them behave according to app timezone based on app configuration, not attempt to store time zone data in DynamoDB itself.

andrykonchin:
@richardhsu 
> So does this occur if you configure Dynamoid to use a different time zone for the dump data?

System time zone is used at dumping and undumping, so `applicatoin_timezone` doesn't affects

@welkie
> They use strings to store dates and datetimes, storing them as an ISO 8601 date locked to UTC:

I assume it would be difficult/impossible to change internal format of type after release so we should make right decision now,  postpone it or even don't support dates at all. Please take into account support of dates is added recently and not released yet.

Storing dates as string is nice and very clear and readable. But there are some cons:
- number value (I mean days count) requires fewer bytes (4 vs 8)  (http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html)
- therefore comparisons of numbers will be cheaper/faster

I haven't made any benchmarks and not sure how it will affect query performance. So I would like to discuss the right way to make benchmark. I am not sure I can take into account all important factors and make fair mesures.

welkie:
@andrykonchin while I was digging around I found internal posts like [this](https://forums.aws.amazon.com/message.jspa?messageID=312432) where they mention that in DynamoDB, numbers are stored as strings, where they use a character for every digit, including the decimal place. This means it would take for example 5 characters to store a day in 2017 if it were a while number representing the number of days since epoch.

I have to admit, that's still much less days used than a Unix timestamp. That would work well Date. For DateTime storing it as a whole number representing the number of milliseconds, or even nanoseconds (like Postgres) since epoch would probably still be less characters than an ISO8601 date string.

One thing that I do feel would be a good idea no matter what is done would be to store the Dates and DateTimes as UTC and allow devs to work with timezones for their apps as they see fit, so that the data in DynamoDB remains objective and clear.

richardhsu:
The only downside is if people need dates before Jan 1, 1970 then where the string format should support theoretically all dates.

welkie:
@richardhsu Would that not be taken care of by storing a negative number? ie.

- *-1* for the day before 1970-01-01
- *-2* for the day before that
etc

richardhsu:
@welkie Ah no you are right, yup we'd be totally fine!

pboling:
So long as we don't alter the existing datetime support which I am currently relying on I think we have some room to wiggle on the new Date format.  I like the idea of the dates plus or minus the Unix epoch.  I also agree that the Date should always be stored in UTC and then converted to the configured time zone on extraction.

lib/dynamoid/persistence.rb
```diff
@@ -135,7 +135,7 @@ def undump_field(value, options)
                 if value.is_a?(Date) || value.is_a?(DateTime) || value.is_a?(Time)
                   value.to_date
                 else
-                  Time.at(value).to_date
+                  Date.new(1970, 1, 1) + value.to_i
```
pboling: Let's make `Date.new(1970, 1, 1)` into a constant.  No reason to instantiate it more than once.

lib/dynamoid/persistence.rb
```diff
@@ -177,7 +177,7 @@ def dump_field(value, options)
             when :datetime
               !value.nil? ? value.to_time.to_f : nil
             when :date
-              !value.nil? ? value.to_time.to_i: nil
+              !value.nil? ? (value.to_date - Date.new(1970, 1, 1)).to_i : nil
```
pboling: Resuse the constant here.

andrykonchin:
@pboling done

PR #180, opened 2017-08-22
AWS SDK V2/Chain: Update query/scan limits & batch

richardhsu:
- [Update] Chain: Changed `eval_limit` to `record_limit` to better
  reflect the limitation on returned records.
- [New] Chain: Added `scan_limit` to allow limiting the number
  of scanned elements and more control over underlying
  query.
- [New] AWS SDK V2: Added `batch_size` option for `query` to
  take advantage of the `batch` method from Chain.
- [Fix] AWS SDK V2: Fix issue where scan could return more
  results than limited. For example having `batch_size` of
  50 and a `record_limit` (was `eval_limit`) of 80 could
  return 100 records since batching would requests in sets
  of 50! Applied similar logic correctly to query.

Cleaning up for https://github.com/Dynamoid/Dynamoid/issues/85 and addressing updates for https://github.com/Dynamoid/Dynamoid/pull/172

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,22 +343,46 @@ u.addresses.where(:city => 'Chicago').all
 
 But keep in mind Dynamoid -- and document-based storage systems in general -- are not drop-in replacements for existing relational databases. The above query does not efficiently perform a conditional join, but instead finds all the user's addresses and naively filters them in Ruby. For large associations this is a performance hit compared to relational database engines.
 
-You can also limit the number of evaluated records, or select a record from which to start, to support pagination:
+#### Limits
+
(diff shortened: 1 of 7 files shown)


andrykonchin:
@richardhsu could you please make separate commit for every feature? It's difficult to review both code and tests in large commit

pboling:
This is no longer needed, right @richardhsu ?

richardhsu:
This is obsolete though I still have one more fix to put in :) Will open new PR since it needed the other two in first otherwise would have been rebasing nightmare.

PR #181, opened 2017-08-22
Rename `eval_limit` -> `record_limit`

richardhsu:
- [Update] Change `eval_limit` to `record_limit` to avoid the stigma
  of `eval` keyword and to be clear when later introducing the
  `scan_limit`.

Part 1 of https://github.com/Dynamoid/Dynamoid/pull/180 with split for rename of `eval_limit` to `record_limit`.

I'll include the new tests from https://github.com/Dynamoid/Dynamoid/pull/180 in the other PR to follow.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,19 +346,21 @@ But keep in mind Dynamoid -- and document-based storage systems in general -- ar
 You can also limit the number of evaluated records, or select a record from which to start, to support pagination:
 
 ```ruby
-Address.eval_limit(5).start(address) # Only 5 addresses.
+Address.record_limit(5).start(address) # Only 5 addresses.
 ```
 
 For large queries that return many rows, Dynamoid can use AWS' support for requesting documents in batches:
 
 ```ruby
-#Do some maintenance on the entire table without flooding DynamoDB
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -346,19 +346,21 @@ But keep in mind Dynamoid -- and document-based storage systems in general -- ar
...
 ```
 
 For large queries that return many rows, Dynamoid can use AWS' support for requesting documents in batches:
 
 ```ruby
-#Do some maintenance on the entire table without flooding DynamoDB
+# Do some maintenance on the entire table without flooding DynamoDB
 Address.all(batch_size: 100).each { |address| address.do_some_work; sleep(0.01) }
-Address.eval_limit(10_000).batch(100). each { … } #batch specified as part of a chain
+Address.record_limit(10_000).batch(100). each { … } # Batch specified as part of a chain
```
pboling: A space following a `.` and then more code on the same line is not valid syntax.
Should be:
```
Address.record_limit(10_000).batch(100).each { … } # Batch specified as part of a chain
```

richardhsu: Thanks good catch!

PR #182, opened 2017-08-22
AWS SDK V2 and Chain: Add `scan_limit`

richardhsu:
- [Update] Add new method `scan_limit` to allow modifying the number
  of scanned records that DynamoDB looks at. This is different from
  `eval_limit` which is the limit of records returned.

Part 2 of https://github.com/Dynamoid/Dynamoid/pull/180

Will combine the last `batch_size` and fix for bad batching when dealing with limits once part 1 and part 2 are merged.

Reference:
Part 1: https://github.com/Dynamoid/Dynamoid/pull/181

Thanks! I kept the `eval_limit` code somewhat in take and will rebase correctly once Part 1 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,10 +343,29 @@ u.addresses.where(:city => 'Chicago').all
 
 But keep in mind Dynamoid -- and document-based storage systems in general -- are not drop-in replacements for existing relational databases. The above query does not efficiently perform a conditional join, but instead finds all the user's addresses and naively filters them in Ruby. For large associations this is a performance hit compared to relational database engines.
 
-You can also limit the number of evaluated records, or select a record from which to start, to support pagination:
+#### Limits
+
+There are three types of limits that you can query with:
+
+1. `record_limit` - The number of evaluated records that are returned by the query.
+2. `scan_limit` - The number of scanned records that DynamoDB will look at before returning.
+3. `batch_size` - The number of records requested to DynamoDB per underlying request, good for large queries!
+
+Using these in various combinations results in the underlying requests to be made in the smallest size possible and
+the query returns once `record_limit` or `scan_limit` is satisfied. It will attempt to batch whenever possible.
+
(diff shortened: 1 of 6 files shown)


pboling:
@richardhsu ready for rebase!

README.md
```diff
@@ -343,24 +343,46 @@ u.addresses.where(:city => 'Chicago').all
...
 But keep in mind Dynamoid -- and document-based storage systems in general -- are not drop-in replacements for existing relational databases. The above query does not efficiently perform a conditional join, but instead finds all the user's addresses and naively filters them in Ruby. For large associations this is a performance hit compared to relational database engines.
 
-You can also limit the number of evaluated records, or select a record from which to start, to support pagination:
+#### Limits
+
+There are three types of limits that you can query with:
+
+1. `record_limit` - The number of evaluated records that are returned by the query.
+2. `scan_limit` - The number of scanned records that DynamoDB will look at before returning.
+3. `batch_size` - The number of records requested to DynamoDB per underlying request, good for large queries!
```
pboling: Perhaps we should mention that there is a maximum batch size set my Amazon, currently documented as 100.
```
BatchGetItem
A single BatchGetItem operation can retrieve a maximum of 100 items. The total size of all the items retrieved cannot exceed 16 MB.
```
http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html

richardhsu: The Query and Scan don't operate on this limit, but I think it's worth clarifying, I can add it to my next PR. Not sure if our `BatchGetItem` has this incorporated already.

```
Query
The result set from a Query is limited to 1 MB per call. You can use the LastEvaluatedKey from the query response to retrieve more results.

Scan
The result set from a Scan is limited to 1 MB per call. You can use the LastEvaluatedKey from the scan response to retrieve more results.
```

PR #183, opened 2017-08-24
AWS SDK V2: Add batching to query and fix paging

richardhsu:
- [New] Add batch size to queries as well which should allow
  any Chain querying to use batching correctly
- [Fix] Fix issue where scan could return more
  results than limited. For example having `batch_size` of
  50 and a `record_limit` (was `eval_limit`) of 80 could
  return 100 records since batching would requests in sets
  of 50! Applied similar logic correctly to query.

Part 3 of https://github.com/Dynamoid/Dynamoid/pull/180

Thanks!

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -448,10 +448,11 @@ module Dynamoid
         opts.delete(:select)
         opts.delete(:index_name)
 
-        # Deal with various limits
+        # Deal with various limits and batching
         record_limit = opts.delete(:record_limit)
         scan_limit = opts.delete(:scan_limit)
-        limit = [record_limit, scan_limit].compact.min
+        batch_size = opts.delete(:batch_size)
+        limit = [record_limit, scan_limit, batch_size].compact.min
         q[:limit] = limit if limit
 
         opts.delete(:next_token).tap do |token|
@@ -505,6 +506,15 @@ module Dynamoid
           record_count = 0
           scan_count = 0
           loop do
+            # Adjust the limit down if the remaining record or scan limit are
+            # lower to obey limits. We can assume the difference won't be
+            # negative due to break statements below.
+            if q[:limit] && record_limit && record_limit - record_count < q[:limit]
+              q[:limit] = record_limit - record_count
+            elsif q[:limit] && scan_limit && scan_limit - scan_count < q[:limit]
+              q[:limit] = scan_limit - scan_count
+            end
+
             results = client.query(q)
             results.items.each { |row| y << result_item_to_hash(row) }
 
@@ -541,9 +551,9 @@ module Dynamoid
         # Deal with various limits and batching
         record_limit = select_opts.delete(:record_limit)
         scan_limit = select_opts.delete(:scan_limit)
-        request_limit = [record_limit, scan_limit].compact.min
-        batch = select_opts.delete(:batch_size)
-        request[:limit] = batch || request_limit if batch || request_limit
+        batch_size = select_opts.delete(:batch_size)
+        request_limit = [record_limit, scan_limit, batch_size].compact.min
+        request[:limit] = request_limit if request_limit
 
         if scan_hash.present?
           request[:scan_filter] = scan_hash.reduce({}) do |memo, (attr, cond)|
@@ -559,6 +569,15 @@ module Dynamoid
           record_count = 0
           scan_count = 0
           loop do
+            # Adjust the limit down if the remaining record or scan limit are
+            # lower to obey limits. We can assume the difference won't be
+            # negative due to break statements below.
+            if request[:limit] && record_limit && record_limit - record_count < request[:limit]
+              request[:limit] = record_limit - record_count
+            elsif request[:limit] && scan_limit && scan_limit - scan_count < request[:limit]
+              request[:limit] = scan_limit - scan_count
+            end
+
             results = client.scan(request)
             results.items.each { |row| y << result_item_to_hash(row) }
 
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -506,12 +506,14 @@ module Dynamoid
           record_count = 0
           scan_count = 0
           loop do
-            # Adjust the limit down if the remaining record or scan limit are
+            # Adjust the limit down if the remaining record and/or scan limit are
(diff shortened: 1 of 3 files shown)


lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -505,6 +506,15 @@ def query(table_name, opts = {})
...
           loop do
+            # Adjust the limit down if the remaining record or scan limit are
+            # lower to obey limits. We can assume the difference won't be
+            # negative due to break statements below.
+            if q[:limit] && record_limit && record_limit - record_count < q[:limit]
+              q[:limit] = record_limit - record_count
+            elsif q[:limit] && scan_limit && scan_limit - scan_count < q[:limit]
+              q[:limit] = scan_limit - scan_count
+            end
+
```
andrykonchin: I beleive we should honor `scan_limit` as well as `record_limit` (it was described [here ](https://github.com/Dynamoid/Dynamoid/pull/172#discussion_r133783235)). So I can imagine case when `record_limit - record_count` is greater than `scan_limit - scan_count`. That's why I think both conditions should be separated. I mean

```ruby
            if q[:limit] && record_limit && record_limit - record_count < q[:limit]
               q[:limit] = record_limit - record_count
            end
            if q[:limit] && scan_limit && scan_limit - scan_count < q[:limit]
              q[:limit] = scan_limit - scan_count
            end

```

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -559,6 +569,15 @@ def scan(table_name, scan_hash, select_opts = {})
...
           loop do
+            # Adjust the limit down if the remaining record or scan limit are
+            # lower to obey limits. We can assume the difference won't be
+            # negative due to break statements below.
+            if request[:limit] && record_limit && record_limit - record_count < request[:limit]
+              request[:limit] = record_limit - record_count
+            elsif request[:limit] && scan_limit && scan_limit - scan_count < request[:limit]
+              request[:limit] = scan_limit - scan_count
+            end
+
```
andrykonchin: There is a case when this approach works not well. I don't know how to solve the issue now but we should remember  that it exists.

We tune limit according to already loaded items count and can appear in a situation when `request[:limit]` is very small in comparison to items count we need to scan to load all requested items (in case when user specified some conditions) and we will make a lot of excessive HTTP calls. For instance `request[:limit]` is 1 but nearest matched item is at 100th position. So we will make 100 HTTP calls.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
richardhsu: Ah good call! Will make that change, thanks!

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
richardhsu: I don't think this is the case with `limit: 1` DynamoDB will only return at most 1 record but it doesn't mean it won't underlying scan through 100 unmatched items. I think this type of case arises with `scan_limit` being set but to which case you likely hit `scan_limit` before your `record_limit` is satisfied then.

The only related edge case which I noticed during testing (with DynamoDBLocal) but not sure if it's true in production is that SCAN will only scan through maximum of 1MB of data and then return so if your record sizes are large (Forget maybe 400KB is max record size?) then if you are requesting for records that total 1GB then you potentially are making a lot of HTTP calls to request all the items since each response will contain the maximum 1MB but that's a fault of DynamoDB response system and not really anything we can do about it.

The other case I can think of is making 100 HTTP calls would happen only if there are 100 items and the user requests to `batch(1)` which that's by design of the programming.

andrykonchin: >  SCAN will only scan through maximum of 1MB of data

it's documented behaviour - http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Pagination

> I don't think this is the case with limit: 1

To be clear I meant `record_limit`. Okay. Let's consider an example.

User sets `record_limit` and some conditions, like `User.where('created_at.gt' => Time.now - 1.day).record_limit(1000)`

First api call - `request[:limit]` is 1000
1000 items were scanned and only 999 were filtered and returned

Next api call - `request[:limit]` is 1000-999=1

So we will make several api calls with this small limit (1) until find one matched item

richardhsu: > SCAN will only scan through maximum of 1MB of data

Hm, I don't see the statement you've quoted. Most statements can be interpreted as the resulting set has maximum of 1 MB. Nothing about the amount of data that is looked at in underlying query/scan.

I have a large query example in which there are 200 records generated and 64 records will be 1 MB in size.

When doing a query with scan limit of 100. The query will return `scanned_count` as 100. But when run with a scan, it returns `scanned_count` as 64. Which shows the discrepancy (but I've noticed there are discrepancies with the local dynamodb jar and how production actual works) 

## Record Limit

You are correct! I kept thinking that the limit past was the limit of returned records, this is **not** correct. The limit given to the SDK is the scanned limit (number of records dynamodb looks at when evaluating criteria).

I will add the note as you suggested. So we have the trade off of:

* Either make small sets of HTTP calls to get the last remaining records
* Do NOT scale down limit when `record_count` approaches `record_limit` and instead during enumeration, stop once we've hit the limit. This will thus load `limit - remaining_records_needed` extra records. So in your example we'd actually be fine. But if there were 1000 extra records satisfying the condition then we'd have 999 extra records loaded into memory.

I don't see any other way to solve this since you basically the in between would be requesting in smaller chunks but you're just moving the needle from calling fewer HTTP calls or more HTTP calls.

richardhsu:
@andrykonchin added note on the example given, let me know if you have any feedback on it! Thanks!

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
andrykonchin: > Hm, I don't see the statement you've quoted. Most statements can be interpreted as the resulting set has maximum of 1 MB. Nothing about the amount of data that is looked at in underlying query/scan.

Just FYI - there is a clean statement in provided article:

"A Query operation can retrieve a maximum of 1 MB of data. This limit applies before the filter expression is evaluated."

andrykonchin:
👍

andrykonchin:
@pboling Looks like nothing stops us to release new version. Maybe after some internal tests

PR #185, opened 2017-08-25
Fix STI support

andrykonchin:
`where`, finders and friends take into account STI (single table inheritence) now

Was implemented STI like it works in Rails' ActiveRecord - query will return items of the model class and all his subclasses

### DIFF
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -780,6 +780,27 @@ describe Dynamoid::Criteria::Chain do
     end
   end
 
+
+  context 'single table inheritance' do
+    describe 'where' do
+      it 'honors STI' do
+        Vehicle.create(description: 'Description')
+        car = Car.create(description: 'Description')
+
+        expect(Car.where(description: 'Description').all).to eq [car]
+      end
+    end
+
+    describe 'all' do
+      it 'honors STI' do
+        Vehicle.create(description: 'Description')
+        car = Car.create
+
+        expect(Car.all).to eq [car]
+      end
+    end
+  end
+
   describe 'User' do
     let(:chain) { described_class.new(User) }
 
--- a/spec/dynamoid/persistence_spec.rb
+++ b/spec/dynamoid/persistence_spec.rb
@@ -542,6 +542,7 @@ describe Dynamoid::Persistence do
   end
 
   context 'single table inheritance' do
+    let(:vehicle) { Vehicle.create }
     let(:car) { Car.create(power_locks: false) }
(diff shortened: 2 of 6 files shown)


spec/dynamoid/criteria/chain_spec.rb
```diff
@@ -769,6 +769,27 @@ def request_params
...
+    end
+
+    describe 'all' do
+      it 'honors STI' do
+        Vehicle.create(description: 'Description')
+        car = Car.create
+
+        expect(Car.all).to eq [car]
+      end
+    end
```
richardhsu: Also add the test where find `Car.all` and make sure it is including all subclass objects so create a `Car` and `Cadillac`.

spec/dynamoid/persistence_spec.rb
```diff
@@ -559,6 +560,13 @@ def reload_address
         expect(v).to include(s)
       }
     end
+
+    it 'does not load parent item when quering the child table' do
+      vehicle && car
+
+      expect(Car.all).to contain_exactly(car)
+      expect(Car.all).not_to include(vehicle)
```
richardhsu: Also make sure this doesn't load `sub` in the set too since different type.

spec/dynamoid/criteria/chain_spec.rb
andrykonchin: @richardhsu 
I am not sure
We already have such test case here https://github.com/Dynamoid/Dynamoid/pull/185/files#diff-686b0e45ce18ac9c53508980fafe31a0R554

TBH I dislike how test cases are organized and grouped with contexts and files in this project. It's difficult for me to find the right place for new test and to find a place where some functionality is already testing

richardhsu: Yeah, I agree, it's difficult. And thanks, I missed that test, I was searching through them to see and guess I just happened to miss that one! Looks good then!

spec/dynamoid/persistence_spec.rb
andrykonchin: :+1: will update PR on Monday

pboling:
@andrykonchin do we want this PR in the release, or not?  Build failure looks like a flaky spec for a single ruby engine.

andrykonchin:
@pboling 

> do we want this PR in the release, or not?

I wouldn't like to release such "fresh" fix. It isn't a critical bug and I think it could wait for the next release

> Are you still going to update this PR more before it is ready?

This PR is ready

pboling:
Dynamoid 1.3.4 is released.  This should be rebased / build fixed and merged!

andrykonchin:
done

PR #186, opened 2017-08-25
Fix query result lazy loading

andrykonchin:
Old behaviour of `Model.where({}).all`:
- load lazily if user specified batch size
- load all collection into memory otherwise

Proposed behaviour - always return lazy evaluated collection

It means `Model.where({}).all` returns `Enumerator` instead of `Array`. If you need `Array` interface you have to convert collection to `Array` manually with `to_a` method call

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -56,9 +56,11 @@ module Dynamoid #:nodoc:
       end
 
       # Returns the last fetched record matched the criteria
+      # Enumerable doesn't implement it, only `first`
+      # So we have to implement it themselves
       #
       def last
-        all.last
+        all.to_a.last
       end
 
       # Destroys all the records matching the criteria.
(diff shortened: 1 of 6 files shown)


andrykonchin:
@pboling @richardhsu I propose to not merge open PRs untill new release. I beleive we should release new version with bug fixes ASAP

pboling:
I agree, this will be merged after the new release.

lib/dynamoid/criteria/chain.rb
```diff
@@ -56,9 +56,11 @@ def all
       end
 
       # Returns the last fetched record matched the criteria
+      # Enumerable doesn't implement it, only `first`
+      # So we have to implement it themselves
```
pboling: `it` => `last` (more specific)
`themselves` => `ourselves` (grammar)

andrykonchin:
@pboling done

pboling:
Dynamoid 1.3.4 is released.  This should be rebased / build fixed and merged!

andrykonchin:
done

pboling:
I have created a [1-3-stable branch](https://github.com/Dynamoid/Dynamoid/blob/1-3-stable/README.md) for maintenance of the old version (current release), so now we can start working on the next version, which I expect will have major breaking changes, and so should be version 2.0.0.

PR #188, opened 2017-08-30
Chain: Fix batch_size for queries

richardhsu:
- [Update] Queries weren't passing the `batch_size` to the underlying
  adapter. Add and added test to test for this behavior.
- [Update] Modified changelog in previous commit accidentally with
  blanket `eval_limit` replaced with `record_limit`. Changing back
  since that statement is not correct in the changelog.

Related Issue https://github.com/Dynamoid/Dynamoid/issues/85

### DIFF
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -81,14 +81,25 @@ describe Dynamoid::Criteria::Chain do
         expect(model.where(request_params.merge({ name: 'Josh' })).batch(3).count).to eq(10)
       end
 
-      it 'supports combined limits' do
+      it 'supports combined limits with batch size 1' do
         # Scanning through 13 means it'll see 10 Josh objects and then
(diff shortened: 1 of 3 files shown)


spec/dynamoid/criteria/chain_spec.rb
```diff
@@ -81,15 +81,26 @@ def request_params
...
       end
 
-      it 'supports combined limits' do
+      it 'supports combined limits with batch size 1' do
         # Scanning through 13 means it'll see 10 Josh objects and then
-        # 3 Pascal objects to which it'll return 2 since the record_limit will
-        # restrict it
+        # 3 Pascal objects but it'll hit record_limit first with 2 objects
+        # so we'd only see 12 requests due to batching.
+        expect(Dynamoid.adapter.client).to receive(request_type).exactly(12).times.and_call_original
```
andrykonchin: Do we test batch loading for `Scan`? Adapter is tested well but chain is not. We need at least smoky tests for limits and batch loading

PR #191, opened 2017-09-13
Field Defaults: Allow default primitive w/o Lambda

richardhsu:
- [Update] Persistence was expecting the default value to support
  either `call` or `dup` but certain values like numbers and booleans
  do not support `dup` since they are copy by value. Adding support
  so can write `:default => 1` or `{ default: 1 }` as README
  suggests with `field :actions_taken, :integer, {default: 0}`.

CR @pboling @andrykonchin thanks!

### DIFF
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -229,6 +229,20 @@ module Dynamoid
           val
         end
       end
+
+      # Evaluates the default value given, this is used by undump
+      # when determining the value of the default given for a field options.
+      #
+      # @param [Object] :value the attribute's default value
+      def undump_default_value(val)
+        if val.respond_to?(:call)
+          val.call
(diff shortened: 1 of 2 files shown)


pboling:
We should start adding to a `HEAD` section of a changelog with every PR so that we don't have so much work to do when preparing for the next release.  Determine if the changes are breaking, or backwards compatible, new feature or bug fix, and add a note detailing the change to the CHANGELOG.

richardhsu:
Oh, yes sounds good! Will do that moving forward! :)

andrykonchin:
@pboling please wait for my review before merging. I need at least a day to look at PR because of time zone shift

andrykonchin:
@richardhsu 👍

pboling:
OK, will wait in the future.

PR #195, opened 2017-09-25
Change behaviour of `find`

andrykonchin:
It's common to get exception if one or more records can not be found for the requested ids

Both `Mongoid` and `ActiveRecord` support this behaviour ([here](https://mongoid.github.io/old/en/mongoid/docs/querying.html) and [here](http://api.rubyonrails.org/v5.1/classes/ActiveRecord/FinderMethods.html#method-i-find))

Current behaviour - `find` returns `nil` or smaller array.
New behaviour - it raises `RecordNotFound` if one or more records can not be found for the requested ids

### DIFF
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -107,7 +107,7 @@ module Dynamoid #:nodoc:
       def exists?(id_or_conditions = {})
         case id_or_conditions
           when Hash then where(id_or_conditions).first.present?
-          else !! find(id_or_conditions)
+          else !! find_by_id(id_or_conditions)
         end
       end
 
--- a/lib/dynamoid/finders.rb
+++ b/lib/dynamoid/finders.rb
@@ -36,9 +36,12 @@ module Dynamoid
         ids = Array(ids.flatten.uniq)
         if ids.count == 1
           result = self.find_by_id(ids.first, options)
+          raise Errors::RecordNotFound if result.nil?
           expects_array ? Array(result) : result
         else
-          find_all(ids)
+          result = find_all(ids)
+          raise Errors::RecordNotFound if result.size != ids.size
+          result
         end
       end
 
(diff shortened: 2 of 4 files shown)


lib/dynamoid/finders.rb
```diff
@@ -36,9 +36,12 @@ def find(*ids)
         ids = Array(ids.flatten.uniq)
         if ids.count == 1
           result = self.find_by_id(ids.first, options)
+          raise Errors::RecordNotFound if result.nil?
           expects_array ? Array(result) : result
         else
-          find_all(ids)
+          result = find_all(ids)
+          raise Errors::RecordNotFound if result.size != ids.size
```
richardhsu: Do we want to provide the error message too?

ActiveRecord will say which id was passed and for an array of ids it'll say which ids passed and it'll produce "found X results, but was looking for N" kind of message so we could help report the difference.

andrykonchin: @richardhsu Done

lib/dynamoid/document.rb
```diff
@@ -107,7 +107,7 @@ def build(attrs = {})
       def exists?(id_or_conditions = {})
         case id_or_conditions
           when Hash then where(id_or_conditions).first.present?
-          else !! find(id_or_conditions)
+          else !! find_by_id(id_or_conditions)
```
pboling: Would like to see a spec for any behavior change we expect from switching to `find_by_id`.  This looks like it might be a significant change.

andrykonchin: @pboling `find` uses `find_by_id` internally. The only difference between them - `find` supports passing multiple ids. As far as I see `Document.exists?` accepts either one id or condition.

So `Document.exists?([1, 2, 3])` will not work. Do you think we should keep this behaviour?

andrykonchin: @pboling Looks like neither ActiveRecord nor Mongoid supports `exists?(array)` signature ([here](https://apidock.com/rails/ActiveRecord/FinderMethods/exists%3F) and [here](https://mongoid.github.io/old/en/mongoid/docs/querying.html)) and user can check multiple ids in the following way
`Document.exists?('id.in': [1, 2, 3])`

pboling: Ok, I think the change is fine then.  Seems we did not have a spec for the old behavior of passing an array directly then.

PR #196, opened 2017-09-25
Change `save`'s return value

andrykonchin:
Current behaviour - `save` returns `self` if model is saved successfully
New behaviour - it returns `true`

Both `Mongoid` and `Active Record` supports this behaviour ([here](http://www.rubydoc.info/github/mongoid/mongoid/Mongoid/Persistable/Savable) )

### DIFF
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -275,8 +275,6 @@ module Dynamoid
       else
         persist
       end
-
-      self
     end
 
     #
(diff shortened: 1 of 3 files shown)


andrykonchin:
@richardhsu 

> Do we care to be in parity with ActiveRecord saving

I believe we should

> For new object chained it is but for new object later saved it is not

Sorry, don't understand
`model.save` never raises exceptions. There is a test case for this behaviour [here](https://github.com/Dynamoid/Dynamoid/pull/196/files#diff-6cd75ac46082b40962e40cabcad62e4fR86)

PR #198, opened 2017-10-05
Create multiple objects

andrykonchin:
Add support of multiple documents creation with `.create` and `.create!`

### DIFF
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -87,7 +87,11 @@ module Dynamoid #:nodoc:
       #
       # @since 0.2.0
       def create!(attrs = {})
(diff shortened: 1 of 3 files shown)


lib/dynamoid/document.rb
```diff
@@ -83,7 +87,11 @@ def create(attrs = {})
       #
       # @since 0.2.0
       def create!(attrs = {})
-        build(attrs).tap(&:save!)
+        if attrs.is_a?(Array)
+          attrs.map { |attr| create!(attr) }
+        else
+          build(attrs).tap(&:save!)
+        end
```
pboling: Create is defined the same way, 2x.  I think we need to just have one.

PR #199, opened 2017-10-05
Add efficient creation of multiple documents

andrykonchin:
Add `Document.import` method to efficiently create multiple documents

`BatchWriteItem` ([documentation]( http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#batch_write_item-instance_method)) is used so we make fewer HTTP calls

Validation and callbacks are skipped for simplicity. `UnprocessedItems` and `ValidationException` are not handled too.

Example:
```
addresses = Address.import([{city: 'Chicago'}, {city: 'New York'}])
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,13 @@ address.city = 'Chicago'
 address.save
 ```
 
+There is an efficient and low-level way to create multiple documents
+(without validation and callbacks running):
+
+```ruby
+users = User.import([{name: 'Josh'}, {name: 'Nick'}])
+```
+
 ### Querying
 
 Querying can be done in one of three ways:
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -205,6 +205,21 @@ module Dynamoid
         end
       end
 
+      def import(objects)
+        documents = objects.map { |attrs|
+          self.build(attrs).tap { |item|
+            item.hash_key = SecureRandom.uuid if item.hash_key.blank?
+          }
+        }
+
+        documents.each_slice(25) do |docs|
+          Dynamoid.adapter.batch_write_item(self.table_name, docs.map(&:dump))
+        end
+
+        documents.each { |d| d.new_record = false }
+        documents
+      end
+
       private
 
       def undump_hash(hash)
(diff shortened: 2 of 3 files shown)


lib/dynamoid/persistence.rb
```diff
@@ -372,7 +387,7 @@ def dump_field(value, options)
     # @since 0.2.0
     def persist(conditions = nil)
       run_callbacks(:save) do
-        self.hash_key = SecureRandom.uuid if self.hash_key.nil? || self.hash_key.blank?
+        self.hash_key = SecureRandom.uuid if self.hash_key.blank?
```
pboling: Thanks for catching that!

spec/dynamoid/persistence_spec.rb
```diff
@@ -722,4 +722,60 @@ def self.name; 'Doc'; end
...
+    it 'makes batch operation' do
+      expect(Dynamoid.adapter.client).to receive(:batch_write_item).exactly(1).times.and_call_original
+      Address.import([{city: 'Chicago'}, {city: 'New York'}])
+    end
+
+    it 'makes multiple batch operations if the limit (25 items) exceeded' do
+      expect(Dynamoid.adapter.client).to receive(:batch_write_item).exactly(2).times.and_call_original
+      Address.import(Array.new(26, {city: 'Chicago'}))
+    end
+  end
```
pboling: Great specs

README.md
```diff
@@ -339,6 +339,13 @@ address.city = 'Chicago'
 address.save
 ```
 
+There is an efficient and low-level way to create multiple documents
+(without validation and callbacks running):
+
+```ruby
+users = User.import([{name: 'Josh'}, {name: 'Nick'}])
+```
+
```
pboling: Very nice!

latortuga:
Any chance of a release with this feature? Could really use it 😄 👍 💯

pboling:
@andrykonchin Can you add this to the changelog, and specify if it is a breaking change.  Following [semver 2.0](http://semver.org/), what should we release next?

andrykonchin:
@pboling done

It isn't a breaking change. Just new method was added

> Following semver 2.0, what should we release next?
Do you mean which version?

TBH I don't know. Formally we have break changes (`find` and `save` behavior). But they are so minor and could be even considered as fixing to correspond to `Rails`/`Mongoid` default behavior

pboling:
We should release a minor point update then.  I will release.

PR #202, opened 2017-10-31
Fix associations

andrykonchin:
Fix several issues with single associations:

* setting `nil` value raises an exception
* document doesn't keep assigned model and loads it from the storage
* `delete` call doesn't clean update cached ids of associated models

### DIFF
--- a/dynamoid.gemspec
+++ b/dynamoid.gemspec
@@ -50,7 +50,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency(%q<activesupport>, [">= 4"])
   spec.add_runtime_dependency(%q<aws-sdk-resources>, ["~> 2"])
   spec.add_runtime_dependency(%q<concurrent-ruby>, [">= 1.0"])
-  spec.add_development_dependency "pry-byebug"
+  spec.add_development_dependency "pry"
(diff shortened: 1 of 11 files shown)


dynamoid.gemspec
```diff
@@ -50,13 +50,12 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency(%q<activesupport>, [">= 4"])
   spec.add_runtime_dependency(%q<aws-sdk-resources>, ["~> 2"])
   spec.add_runtime_dependency(%q<concurrent-ruby>, [">= 1.0"])
-  spec.add_development_dependency "pry", "~> 0.10"
+  spec.add_development_dependency "pry-byebug"
```
richardhsu: Seems like `jruby` isn't happy with `byebug`

spec/dynamoid/associations/belongs_to_spec.rb
```diff
@@ -62,4 +62,132 @@
       expect(user.monthly).to eq subscription
     end
   end
+
+  describe "{association}=" do
```
richardhsu: Use single quotes for the description strings unless there are escaped characters.

andrykonchin:
@richardhsu done

PR #203, opened 2017-11-02
Updated to AWS SDK 3 and specified core and dynamodb gems

jamtur01:


jamtur01:
I am not sure whether this is the right approach with the adapter... Happy to take suggestions. The test suite passes for me with this.

pboling:
This change should be a major version bump for the gem, so will hold one merging until have the next point release *released*, or until we start a `1.3-stable` branch .

jamtur01:
Yeah I presumed so but didn't want to mess with your versioning.

andrykonchin:
Thank you for your help.

This PR is too outdated so it's easier to make changes from scratch.

Closed in favor of https://github.com/Dynamoid/Dynamoid/pull/267

pboling:
Next release of the gem needs to be a major version bump.  People may need to upgrade their SDK in concert with this upgrade.  As the SDK is not fully backwards compatible we should allow that dependency to update without also bumping our major version.

andrykonchin:
@pboling Yep, completely agree. That's why I want to release as many breaking changes as I can prepare

PR #205, opened 2017-11-03
Fix Chain#destroy_all

andrykonchin:
There are following fixes of `delete_all`:
- exception when makes scan and sort key is declared in model
- exception when makes scan and any condition is specified in `where` clause (like `Document.where().delete_all`)
- exception when makes query and sort key isn't declared in model

Updated Readme with examples of `#delete_all` and declaration of global secondary index.

There was a naive implementation of batch deletion with api call for every item. Replaced with real batch deletion via `batch_write_item` api call

Renamed `Chain#destroy_all` as `#delete_all` and kept `destroy_all` available as alias. In Rails `destroy` method destroys the records matching conditions by instantiating each record and calling its destroy method. Callbacks are running too. `delete` just removes a record from a table. So as far as we don't run callbacks we deletes and don't destroy. Actually we instantiate models but we can just load keys and avoid some excessive traffic.


spec/spec_helper.rb
```diff
@@ -11,6 +11,7 @@
 require "byebug" if ENV["DEBUG"]
 
 require "dynamodb_local"
+require "new_class_helper"
```
pboling: See line 37/38 below:
```
 Dir["#{File.dirname(__FILE__)}/support/**/*.rb"].each {|f| require f}
```
All we have to do is put support files in `spec/support` and they will be required.  Please move the `new_class_helper` there.

andrykonchin: @pboling done

lib/dynamoid/criteria/chain.rb
```diff
@@ -70,25 +70,27 @@ def last
...
         if key_present?
-          ranges = []
           Dynamoid.adapter.query(source.table_name, range_query).collect do |hash|
             ids << hash[source.hash_key.to_sym]
-            ranges << hash[source.range_key.to_sym]
+            ranges << hash[source.range_key.to_sym] if source.range_key
           end
 
-          Dynamoid.adapter.delete(source.table_name, ids,{:range_key => ranges})
+          Dynamoid.adapter.delete(source.table_name, ids, range_key: ranges.presence)
```
richardhsu: Oh nice, learned something new today, didn't realize `.presence` was a thing :) actually kind of handy!

spec/dynamoid/criteria/chain_spec.rb
```diff
@@ -819,6 +819,157 @@ def request_params
     end
   end
 
+  describe '#delete_all' do
+    it "deletes in batch" do
```
richardhsu: Single quotes please, unless have escapes

andrykonchin: @richardhsu done

lib/dynamoid/adapter.rb
```diff
@@ -108,7 +108,9 @@ def delete(table, ids, options = {})
           ids = range_key ? [[ids, range_key]] : ids
         end
 
-        batch_delete_item(table => ids)
+        ids.each_slice(25) do |ary|
+          batch_delete_item(table => ary)
```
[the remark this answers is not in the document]
andrykonchin: @richardhsu
> Can we do the slicing in batch_delete_item? 

agree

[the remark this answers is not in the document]
andrykonchin: @richardhsu done

PR #207, opened 2017-11-06
Refactor `batch_write_item`

andrykonchin:
* Added slicing by 25 requests in `#batch_write_item` like it does in `#batch_delete_item`
* Fixed `#delete` method for case `adapter.delete(table_name, [1, 2, 3], range_key: 1)`


andrykonchin:
@pboling Looks like specs passed successfuly. Is it OK for you?

pboling:
Very good!

Issue #135, opened 2017-03-24
boolean field in where() condition


kekla:
In lib/dynamoid/persistence.rb line 317:
`!value.nil? ? value.to_s[0] : nil`
If I save a boolean field as true or false, it will be saved as string 't' or 'f'.
`.where(:some_boolean_field => true)` won't get correct results, I have to change to  `.where(:some_boolean_field => 't')`.
Do you think it's a good idea to use true/false instead?

pboling:
That kind of translation can be done in the adapter (that's how Rails / ActiveRecord does it).
We can't change the way the data is stored at this point as it would break compatibility with old data.  For now you will have to use `t` or `f` in the query.  We should look into translating `true` and `false` in the adapter.

Leaving this open for that.

pboling:
I have implemented this feature in release 1.3.1 just now.  :integer, :string, and :boolean fields are now run through the same dump process that happens when values are written to Dynamo.

Result: you can query like `where(admin: true)` now.

baloran:
Hey, not working for me :/ I have in my model a field of boolean type. And in dynamodb a boolean type too. And I can't make query condition on it.

pboling:
@andrykonchin @richardhsu Have any idea about this?  Perhaps we had a regression, or perhaps my implementation was broken, and we need better specs.

baloran:
If you can explain where look to find debug, I can try to look :)

richardhsu:
Sorry didn't have time to look at this until now. @baloran what version of Dynamoid gem are you using? In our current version I added this to our `spec/criteria/chain.rb` and tests pass:

```
  describe 'Boolean support' do
    let(:model) {
      Class.new do
        include Dynamoid::Document
        table name: :user, key: :name
        field :admin, :boolean
      end
    }

    it 'supports filter boolean field' do
      user1 = model.create(name: 'Bob', admin: true)
      user2 = model.create(name: 'Gene', admin: false)
      user3 = model.create(name: 'Tina', admin: false)
      user4 = model.create(name: 'Louise', admin: true)
      user5 = model.create(name: 'Teddy', admin: 't')
      user6 = model.create(name: 'Frond', admin: 'f')


      expect(model.where(admin: true).all).to contain_exactly(user1, user4, user5)
      expect(model.where(admin: 't').all).to contain_exactly(user1, user4, user5)
    end
  end
```

```
bundle exec rspec spec/dynamoid/criteria/chain_spec.rb:330
[...]

PR #213, opened 2017-12-12
Support native boolean type

andrykonchin:
Related issue https://github.com/Dynamoid/Dynamoid/issues/135

Add `store_as_boolean` option to store boolean field as `true`/`false` instead of `"t"`/`"f"`

Example

```ruby
field :active, :boolean, store_as_native_boolean: true
```

It isn't a breaking change and default behaviour isn't changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,13 +143,13 @@ The primary use case for using a custom field type is to represent your business
 The boolean fields are stored as `"t", "f"` strings by default. DynamoDB
 supports boolean type natively. So if you want to use native boolean
 type or already have table with native boolean attribute you can easily
-acheive this with `store_as_boolean` option:
+achieve this with `store_as_native_boolean` option:
 
 ```ruby
 class Document
   include DynamoId::Document
 
-  field :active, :boolean, store_as_boolean: true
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -138,6 +138,23 @@ By default, fields are assumed to be of type ```:string```. Other built-in types
 If built-in types do not suit you, you can use a custom field type represented by an arbitrary class, provided that the class supports a compatible serialization interface.
 The primary use case for using a custom field type is to represent your business logic with high-level types, while ensuring portability or backward-compatibility of the serialized representation.
 
+#### Note on boolean type
+
+The boolean fields are stored as `"t", "f"` strings by default. DynamoDB
+supports boolean type natively. So if you want to use native boolean
+type or already have table with native boolean attribute you can easily
+acheive this with `store_as_boolean` option:
```
richardhsu: Typo `acheive` -> `acheive`.

Also could we change the naming to `store_as_native_boolean: true` or shorter `native_boolean: true` That might call more to action that it's the underlying DynamoDB layer of storage?

pboling: That was my typo :( . Thanks for fixing!

PR #215, opened 2017-12-19
Update CHANGELOG for 2.0.0 release

pboling:
Would like to release 2.0.0 tonight.


pboling:
It is important to communicate the risks of upgrading to the users though, and semantic versioning is the standard way of doing that.  We can either think hard about maintaining a deprecated "API" serviceable that works with the old way when we make otherwise breaking changes, or we can bump the version.

The rake gem often bumps the major version, and some other gems do as well.  If the gem is undergoing upheaval it is important to communicate.  There isn't too much value in keeping the same version number.  We want people to conscientously upgrade, and with a major version bump they are more likely to check the changelog.  It will also make the versioning work automatically with dependency management services like [depfu](https://depfu.com/).

andrykonchin:
I suppose the easiest change will be ready in a month so it doesn't make sense to wait for me.
Let's release now.

PR #216, opened 2017-12-31
Fix global index detection

andrykonchin:
Fix issue with incorrect using of global secondary index.

```ruby
class Message
  include Dynamoid::Document

  table key: :guid
  field :message_thread_guid

  global_secondary_index hash_key: :message_thread_guid, projected_attributes: :all
end
```

Querying with condition other than "equal" for hash-key of some index 

```ruby
Message.where('message_thread_guid.in' => ['1', '2']).to_a
```
leads to "Aws::DynamoDB::Errors::ValidationException: One or more parameter values were invalid: Condition parameter type does not match schema type" error

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -294,7 +294,7 @@ module Dynamoid #:nodoc:
         # But only do so if projects ALL attributes otherwise we won't
         # get back full data
         source.global_secondary_indexes.each do |_, gsi|
-          next unless query_keys.include?(gsi.hash_key.to_s) && gsi.projected_attributes == :all
+          next unless query.keys.map(&:to_s).include?(gsi.hash_key.to_s) && gsi.projected_attributes == :all
           @hash_key = gsi.hash_key
           @range_key = gsi.range_key
           @index_name = gsi.name
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -604,6 +604,15 @@ describe Dynamoid::Criteria::Chain do
         expect(chain).to receive(:records_via_query).and_call_original
         expect(chain.where(city: 'San Francisco').start(@customer2).all).to contain_exactly(@customer3)
       end
+
+      it "does not use index if a condition for index hash key is other than 'equal'" do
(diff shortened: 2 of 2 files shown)


lib/dynamoid/criteria/chain.rb
```diff
@@ -294,7 +294,7 @@ def key_present?
         # But only do so if projects ALL attributes otherwise we won't
         # get back full data
         source.global_secondary_indexes.each do |_, gsi|
-          next unless query_keys.include?(gsi.hash_key.to_s) && gsi.projected_attributes == :all
+          next unless query.keys.map(&:to_s).include?(gsi.hash_key.to_s) && gsi.projected_attributes == :all
```
pboling: On line 276 above we have this:
```
if query_keys.include?(source.range_key.to_s)
```

Seems like we may need a more complete solution than trying to remember to do this `to_s` conversion everywhere.

andrykonchin: @pboling I considered wrapping `query` hash in OOP stile to incapsulate some logic and hide type conversion but still don't have simple and flexible solution

pboling: This bug is too narrow still.  The problem is that we are comparing hash keys with strings, and the hash keys may not themselves be strings.  Fixing this one instance but leaving the bug in other places where we do comparisons seems like a bad idea to me, as we'll just lose track of the actual problem.  I think we should go with a more encompassing solution.  Doesn't need to be OOP necessarily, but, why not have new method like `query_keys` that gives them as string, e.g. `query_keys_as_string`, and then call that from everywhere we do key comparisons.

andrykonchin: @pboling Agree

The simplest solution (and it's already used here in this project) is to convert all user strings to the same type, e.g. symbols. And to have proper tests :).

I will check how we handle and store all user strings like field names, key names, conditions etc to ensue we convert them to symbols.

Actually this bug isn't related to symbol/string comparison. We need to check `hash_key: 'value'` condition but `query_keys` contains filtered field names without operators so `'hash_key.contains': 'value'` triggers involving index too.

andrykonchin:
@pboling @richardhsu If nobody don't mind I will merge this fix

pboling:
I'm ok with it.

PR #221, opened 2018-01-11
Implement fixed datatype for elements in set

pratik60:
Currently a set of integers stored in a set, returns a set of big decimal numbers.
```
model = Class.new do
  include Dynamoid::Document
  table name: :customer
  field :tag_ids, :set
end

model.create(tag_ids: Set.new([1,2]))

model.tag_ids => A set of Big Decimal numbers
```

This pr allows you to add a set_collection_type, and if integer is specified, it will return a set of integers. 

```
model = Class.new do
  include Dynamoid::Document
  table name: :customer
  field :tag_ids, :set, { of: :integer }
end

model.create(tag_ids: Set.new([1,2]))

model.tag_ids => Set.new(1,2)
```

### DIFF
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -117,7 +117,7 @@ module Dynamoid
                   value
                 end
               when :set
-                Set.new(value)
+                undump_set(options, value)
               when :datetime
                 parse_datetime(value, options)
               when :date
@@ -138,6 +138,17 @@ module Dynamoid
                 raise ArgumentError, "Unknown type #{options[:type]}"
             end
           end
+        end
+      end
+
+      def undump_set(options, value)
+        case options[:of]
+        when :integer
+          value.map { |v| Integer(v) }.to_set
+        when :number
+          value.map { |v| BigDecimal.new(v.to_s) }.to_set
+        else
+          value.is_a?(Set) ? value : Set.new(value)
         end
       end
 
--- a/spec/dynamoid/persistence_spec.rb
+++ b/spec/dynamoid/persistence_spec.rb
@@ -454,6 +454,31 @@ describe Dynamoid::Persistence do
     end
   end
 
+  describe "Set field" do
+    let(:klass) do
+      new_class do
+        field :string_set, :set
+        field :integer_set, :set, { of: :integer }
+        field :number_set, :set, { of: :number }
+      end
+    end
+
+    it "stored a string set" do
+      obj = klass.create(string_set: Set.new(['a','b']))
+      expect(obj.reload[:string_set]).to eq(Set.new(['a','b']))
+    end
+
+    it "stored an integer set" do
+      obj = klass.create(integer_set: Set.new([1,2]))
+      expect(obj.reload[:integer_set]).to eq(Set.new([1,2]))
+    end
+
+    it "stored a number set" do
+      obj = klass.create(number_set: Set.new([1,2]))
+      expect(obj.reload[:number_set]).to eq(Set.new([BigDecimal(1),BigDecimal(2)]))
+    end
+  end
+
   it 'raises on an invalid boolean value' do
     expect do
       address.deliverable = true


lib/dynamoid/persistence.rb
```diff
@@ -152,6 +152,17 @@ def undump_field(value, options)
         end
       end
 
+      def undump_set(options, value)
+        case options[:item_type]
```
andrykonchin: I would call it just `type`.

`item` is used for the whole document in the DynamoDb documentation (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html?shortFooter=true)

lib/dynamoid/persistence.rb
```diff
@@ -152,6 +152,17 @@ def undump_field(value, options)
...
+        when :integer
+          value.map { |v| Integer(v) }.to_set
+        when :number
+          value.map { |v| BigDecimal.new(v.to_s).to_f }.to_set
+        else
+          value
```
andrykonchin: should be `Set.new(value)`

lib/dynamoid/persistence.rb
```diff
@@ -152,6 +152,17 @@ def undump_field(value, options)
         end
       end
 
+      def undump_set(options, value)
+        case options[:item_type]
+        when :integer
+          value.map { |v| Integer(v) }.to_set
+        when :number
+          value.map { |v| BigDecimal.new(v.to_s).to_f }.to_set
```
andrykonchin: Default behavior now is to use `BigDecimal` in order to store set of numbers.
You propose to add new value `number` and to convert to Float. I don't understand why.

For numeric set if you save integer values `{1, 2, 3}`, you will convert them to float at loading `{1.0, 2.0, 3.0}`. It would be surprise to user to get float when it stores set of **numbers**

Anyway, I would like to see proposed changes either in PR's description or issue (to discuss them before implementation)

pboling: I think we should continue to have `:number` map to `BigDecimal`.  If you want to support float, make it a new type:
```
        when :integer
          value.map { |v| Integer(v) }.to_set
        when :number
          value.map { |v| BigDecimal.new(v.to_s) }.to_set
        when :float
          value.map { |v| v.to_s.to_f }.to_set
```

lib/dynamoid/persistence.rb
pratik60: Type conflicts with the main type (set, serialized, raw) etc. I didn't dig into this deeply, but I think the options hashes get merged/initialized, hence type didn't work for me. 

I went with set_collection_type. What do you think?

lib/dynamoid/persistence.rb
pratik60: Done

lib/dynamoid/persistence.rb
pratik60: Done

lib/dynamoid/persistence.rb
```diff
@@ -158,6 +171,8 @@ def dump_field(value, options)
               !value.nil? ? Integer(value) : nil
             when :number
               !value.nil? ? value : nil
+            when :float
+              !value.nil? ? value.to_f : nil
```
andrykonchin: Looks like we didn't consider adding `float` type, only set of float

andrykonchin:
Looks good to me, but I would consider naming of new option `set_collection_type` again.

What would you say about `of`?
```ruby
field :years, :set, of: :integer
```

andrykonchin:
At the end I would like to have all working commits squashed into one as far as we have very few changes

pratik60:
@andrykonchin Squashed the commits, and set_collection_type -> of. I will leave you to merge once happy.

andrykonchin:
I asked you to move this into separate PR if want to add this feature

```
+            when :float
+              !value.nil? ? value.to_f : nil
```

pratik60:
@andrykonchin I misunderstood. Removed implementation of float from this pr, makes sense 👍

andrykonchin:
👍

PR #223, opened 2018-01-11
Add option to store datetimes as ISO-8601 formatted strings

william101:
Currently datetimes are stored in DynamoDB as `:number` (UNIX timestamps)

This pr add the option to store these timestamps as `:string` (ISO 8601)

Reference: 
> https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.DataTypes.html
> Date (as ISO_8601 millisecond-precision string, shifted to UTC)

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,13 +166,13 @@ end
 
 #### Note on datetime type
 
-By default datetime fields are persisted as UNIX timestamps with milisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `convert_date_to_string` to `true`
+By default datetime fields are persisted as UNIX timestamps with milisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_native_string` to `true`
 
 ```ruby
 class Document
   include DynamoId::Document
 
-  field :sent_at, :datetime, convert_date_to_string: true
+  field :sent_at, :datetime, store_as_native_string: true
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -598,7 +598,7 @@ Listed below are all configuration options.
 * `sync_retry_max_times` - when Dynamoid creates or deletes table synchronously it checks for completion specified times. Default is 60 (times). It's a bit over 2 minutes by default
 * `sync_retry_wait_seconds` - time to wait between retries. Default is 2 (seconds)
 * `convert_big_decimal` - if `true` then Dynamoid converts numbers stored in `Hash` in `raw` field to float. Default is `false`
-* `convert_date_to_string` - if `true`, datetimes will be stores as ISO-8601 formatted strings. When `false`, datetimes will be stores as UNIX timestamps with milisecond precission. Defaule is `false`
+* `convert_date_to_string` - if `true`, datetimes will be stores as ISO-8601 formatted strings. When `false`, datetimes will be stores as UNIX timestamps with milisecond precission. Default is `false`
 * `models_dir` - `dynamoid:create_tables` rake task loads DynamoDb models from this directory. Default is `app/models`. In Rails application you should set `./app/models` value
 * `application_timezone` - Dynamoid converts all `datetime` fields to specified time zone when loads data from the storage.
   Acceptable values - `utc`, `local` (to use system time zone) and time zone name e.g. `Eastern Time (US & Canada)`. Default is `local`
--- a/README.md
+++ b/README.md
@@ -164,6 +164,18 @@ class Document
 end
 ```
 
+#### Note on datetime type
+
+By default datetime fields are persisted as UNIX timestamps with milisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `convert_date_to_string` to `true`
+
+```ruby
+class Document
+  include DynamoId::Document
+
+  field :sent_at, :datetime, convert_date_to_string: true
+end
+```
+
 #### Magic Columns
 
 You get magic columns of id (string), created_at (datetime), and updated_at (datetime) for free.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -598,6 +598,7 @@ Listed below are all configuration options.
 * `sync_retry_max_times` - when Dynamoid creates or deletes table synchronously it checks for completion specified times. Default is 60 (times). It's a bit over 2 minutes by default
 * `sync_retry_wait_seconds` - time to wait between retries. Default is 2 (seconds)
 * `convert_big_decimal` - if `true` then Dynamoid converts numbers stored in `Hash` in `raw` field to float. Default is `false`
+* `convert_date_to_string` - if `true`, datetimes will be stores as ISO-8601 formatted strings. When `false`, datetimes will be stores as UNIX timestamps with milisecond precission. Defaule is `false`
```
pratik60: Typo, default

lib/dynamoid/persistence.rb
```diff
@@ -208,7 +197,7 @@ def dynamo_type(type)
               :string
             else
               raise 'unknown type'
-          end
+            end
```
pratik60: Is this intentional?

william101: Just my editor formatting the code. I'll update.

lib/dynamoid/persistence.rb
```diff
@@ -250,6 +239,36 @@ def undump_hash_value(val)
...
+      end
+
+      def parse_datetime(value)
+        return value if value.is_a?(Date) || value.is_a?(DateTime) || value.is_a?(Time)
+
+        case Dynamoid::Config.application_timezone
+          when :utc
+            time_for_timezone(value, 'UTC')
+          when :local
+            time_for_timezone(value, Time.now.getlocal.zone)
```
andrykonchin: `ActiveSupport::TimeZone` works badly with decree time and may loose 1 hour. That's why we used `Time.at(value).to_datetime` to convert time.

lib/dynamoid/persistence.rb
```diff
@@ -119,18 +119,7 @@ def undump_field(value, options)
...
-                  case Dynamoid::Config.application_timezone
-                    when :utc
-                      ActiveSupport::TimeZone['UTC'].at(value).to_datetime
-                    when :local
-                      Time.at(value).to_datetime
-                    when String
-                      ActiveSupport::TimeZone[Dynamoid::Config.application_timezone].at(value).to_datetime
-                  end
-                end
+                parse_datetime(value)
```
andrykonchin: It's a good refactoring. The feature could be done just with one additional line

```ruby
value = DateTime.iso8601(value).to_time.to_i if Dynamoid::Config.convert_date_to_string
```

lib/dynamoid/config.rb
```diff
@@ -26,6 +26,7 @@ module Config
     option :sync_retry_max_times, default: 60 # a bit over 2 minutes
     option :sync_retry_wait_seconds, default: 2
     option :convert_big_decimal, default: false
+    option :convert_date_to_string, default: false
```
andrykonchin: To support existed projects and data we may need to specify this format not globally but for specific model and field (as well as allow to set it globally)

spec/dynamoid/persistence_spec.rb
```diff
@@ -233,27 +233,46 @@ def reload_address
...
+
+      it 'loads time in UTC if config.application_timezone = :utc', convert_date_to_string: :true do
+        time = '2017-06-20 08:00:00 +0300'.to_time
+        user = User.create(last_logged_in_at: time)
+        user = User.find(user.id)
+        expect(user.last_logged_in_at).to eq '2017-06-20 05:00:00 +0000'.to_datetime
+      end
+    end
+
+    include_examples 'dump_datetime_attributes'
```
andrykonchin: I would wrap it with context `context 'when convert_date_to_string == false' do`

spec/dynamoid/persistence_spec.rb
```diff
@@ -233,27 +233,46 @@ def reload_address
...
+      after { Dynamoid::Config.convert_date_to_string = false }
+
+      include_examples 'dump_datetime_attributes'
+
+      it 'dumps datetime as a string' do
+        time = '2017-06-20 08:00:00 +0300'.to_time
+        user = User.create(last_logged_in_at: time)
+        user = User.find(user.id)
+        expect(user.dump[:last_logged_in_at]).to be_a(String)
+        expect(user.dump[:last_logged_in_at]).to eq '2017-06-20T05:00:00+00:00'
```
andrykonchin: The test passes only because it runs in UTC timezone and it fails when local time zone isn't UTC.
You should set `application_timezone='UTC'` for this test case. It's `local` by default

lib/dynamoid/config.rb
pboling: @andrykonchin wouldn't the default value of `false` make it backwards compatible with any existing implementation?

spec/dynamoid/persistence_spec.rb
pboling: Ideally we would have tests for multiple time zones for any feature based on parsing time.

lib/dynamoid/config.rb
andrykonchin: @pboling It would, sure.
But you could want to keep old format for existing tables and new format for new ones. You couldn't use new format if already have data in old one in any table.

spec/dynamoid/persistence_spec.rb
andrykonchin: @pboling Agree, but I didn't find a way to change timezone in tests. We can travel in time with `Timecop` but we cannot change timezone.

lib/dynamoid/config.rb
pboling: Oh, I see.  Agreed, this has to be model-specific per field.

README.md
```diff
@@ -164,6 +164,18 @@ class Document
...
 
+#### Note on datetime type
+
+By default datetime fields are persisted as UNIX timestamps with milisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `convert_date_to_string` to `true`
+
+```ruby
+class Document
+  include DynamoId::Document
+
+  field :sent_at, :datetime, convert_date_to_string: true
```
andrykonchin: We already have similar option for boolean field `store_as_native_boolean`

```ruby
field :active, :boolean, store_as_native_boolean: true
```

It makes sense to name new option in similar way

andrykonchin: Again, it may be useful to apply this new option globally and override it for specific field

pboling: We should follow existing naming patterns. 👍

Issue #220, opened 2018-01-11
Conditionals fail while filtering for set type


pratik60:
It appears that the `scan_query` method implemented in `chain.rb` fails with `ArgumentError: value must be enumerable` when using conditions for a set. It works for an array. 

For example: 
```
 model = Class.new do
         include Dynamoid::Document
         table name: :customer
         field :skill_set, :set
       end
model.create(skill_set: Set.new(['ruby']))
model.where('skill_set.contains' => 'ruby').all.to_a # This will fail with the above error
```
I created a little PR with a failing spec here: #219 to ease the process.

andrykonchin:
Fixed. Please check master branch.

pratik60:
Wohooo! Thanks for this :-)

PR #224, opened 2018-01-11
Disable type conversion for Set and Array condition values

andrykonchin:
Fix `contains` operator for `set` fields

Example:
`Document.where("set.contains" => "foo").all`

Related issue https://github.com/Dynamoid/Dynamoid/issues/220

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -258,6 +258,8 @@ module Dynamoid #:nodoc:
       end
 
       def type_cast_condition_parameter(key, value)
+        return value if [:array, :set].include?(source.attributes[key.to_sym][:type])
+
         if !value.respond_to?(:to_ary)
           source.dump_field(value, source.attributes[key.to_sym])
         else
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -408,6 +408,30 @@ describe Dynamoid::Criteria::Chain do
         .to contain_exactly(customer1, customer3)
     end
 
+    it 'supports contains for set' do
+      klass = new_class do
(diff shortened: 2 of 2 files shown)


lib/dynamoid/criteria/chain.rb
```diff
@@ -258,6 +258,8 @@ def range_query
       end
 
       def type_cast_condition_parameter(key, value)
+        return value if [:array, :set].include?(source.attributes[key.to_sym][:type])
```
richardhsu: Shouldn't this be the `else` case below? So really the `else` case is wrong because the `dump` field won't work correctly since the type is array even though it's acting on individual values of the array.

Just quick glance, almost seems like this really should just always be, right?

```
source.dump_field(value, source.attributes[key.to_sym]
```

andrykonchin: @richardhsu I think it shouldn't

The value may be collection (Set or Array) in following cases:
* `in` operator - `where("field.in" => [1, 2, 3])`
* `between` operator - `where("field.between" => [0, 10])`
* `equal` operator for `set`/`array` - `where(set_field: Set.new([1, 2, 3]))`

We have to handle cases 1, 2 and 3 in different way. For cases 1, 2 we have to convert each element of collection. For case 3  we have convert the whole value (to `set` or `array`)

`else` case handles cases 1, 2. And here I disable/postpone handling case 3 in order to simplify code. It's a complex logic because we have to take into account not only field type but operator. For instance for `set` field operator `equal` expects `set` but `contains` expects type of `set` element - `number` or `string`

pboling: Thanks for the explanation!

richardhsu: Ah perfect, @andrykonchin thanks for explanation!

PR #228, opened 2018-02-04
Store dates as ISO-8601 formatted strings

andrykonchin:
Similar to https://github.com/Dynamoid/Dynamoid/pull/223

* handle `store_as_string` field option for `date` fields
* rename `datetime` option `store_as_native_string` to `store_as_string`
* add global config options `store_datetime_as_string` and `store_date_as_string`

### DIFF
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -255,7 +255,7 @@ module Dynamoid
           ? Dynamoid.config.store_datetime_as_string \
           : options[:store_as_string]
 
-        use_string_format ? value.iso8601 : value.to_time.to_f
+        use_string_format ? value.to_time.iso8601 : value.to_time.to_f
       end
 
       def format_date(value, options)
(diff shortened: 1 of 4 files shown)


lib/dynamoid/persistence.rb
```diff
@@ -251,13 +251,33 @@ def undump_hash_value(val)
...
+
+        use_string_format ? value.to_time.iso8601 : value.to_time.to_f
+      end
+
+      def format_date(value, options)
+        use_string_format = options[:store_as_string].nil? \
+          ? Dynamoid.config.store_date_as_string \
+          : options[:store_as_string]
+
+        unless use_string_format
```
richardhsu: Did you write it this way to optimize for fact that it'll likely be false?

andrykonchin: @richardhsu Yep. I consider numeric format as a common case and string one as an exception.

PR #229, opened 2018-02-06
Support start() on scan and support hashes for both scan and query start()

mnussbaumer:
Refactored start_key on chain.rb  to return correctly exclusive_start_key  instead of next_token, removed the conversion of next_token into exclusive_start_key from both query and scan  methods in the plugin adapter (aws_sdk_v2.rb). Also added the option to pass a hash to start. Added tests for the hash when querying, and added test for both record and hash when scanning with start. Slight change to the readme to mention the possibility of passing a hash

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,8 +431,8 @@ You can thus limit the number of evaluated records, or select a record from whic
 ```ruby
 Address.record_limit(5).start(address) # Only 5 addresses starting at `address`
 ```
-Where `address` is an instance of the model or a hash {the_model_hash_key: 'value', the_model_range_key: 'value'}:
-Keep in mind that if you are passing a hash to `.start()` you need to explicitly define all required keys in it, including range keys and secondary global indexes, if any, when querying and, range keys, if any, when scanning. Otherwise you'll get an `Aws::DynamoDB::Errors::ValidationException` either for `Exclusive Start Key must have same size as table's key schema` or `The provided starting key is invalid`
+Where `address` is an instance of the model or a hash `{the_model_hash_key: 'value', the_model_range_key: 'value'}`:
+Keep in mind that if you are passing a hash to `.start()` you need to explicitly define all required keys in it including range keys, depending on table or secondary indexes signatures, otherwise you'll get an `Aws::DynamoDB::Errors::ValidationException` either for `Exclusive Start Key must have same size as table's key schema` or `The provided starting key is invalid`
 
 If you are potentially running over a large data set and this is especially true when using certain filters, you may
 want to consider limiting the number of scanned records (the number of records DynamoDB infrastructure looks through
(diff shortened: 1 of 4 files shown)


mnussbaumer:
@richardhsu hey, here's the PR, in the end I also added your suggestion for the hash, this way everything works as before, scan works, but you can also pass it a hash instead of a record, please review it and let me know,
thanks

lib/dynamoid/criteria/chain.rb
```diff
@@ -311,20 +311,17 @@ def key_present?
...
         hash_key = @hash_key || source.hash_key
         range_key = @range_key || source.range_key
 
         key = {}
-        key[:hash_key_element] = type_cast_condition_parameter(hash_key, @start.send(hash_key))
-        key[:range_key_element] = type_cast_condition_parameter(range_key, @start.send(range_key)) if range_key
+        key[hash_key] = type_cast_condition_parameter(hash_key, @start.send(hash_key))
+        key[range_key] = type_cast_condition_parameter(range_key, @start.send(range_key)) if range_key
+        # Add table composite keys if they differ from secondary index used composite key
+        key[source.hash_key] = type_cast_condition_parameter(source.hash_key, @start.hash_key) if hash_key != source.hash_key
```
richardhsu: Can probably remove this blank line if you do one liners, kind of prefer we have lines less than 120 characters though, which is why original line had it split into multi-line if statement.

README.md
```diff
@@ -426,11 +426,13 @@ There are three types of limits that you can query with:
 Using these in various combinations results in the underlying requests to be made in the smallest size possible and
 the query returns once `record_limit` or `scan_limit` is satisfied. It will attempt to batch whenever possible.
 
-You can thus limit the number of evaluated records, or select a record from which to start, to support pagination:
+You can thus limit the number of evaluated records, or select a record from which to start in order to support pagination.
 
 ```ruby
 Address.record_limit(5).start(address) # Only 5 addresses starting at `address`
 ```
+Where `address` is an instance of the model or a hash {the_model_hash_key: 'value', the_model_range_key: 'value'}:
```
richardhsu: Surround hash with ticks please so it shows as code.

README.md
```diff
@@ -426,11 +426,13 @@ There are three types of limits that you can query with:
...
+Keep in mind that if you are passing a hash to `.start()` you need to explicitly define all required keys in it, including range keys and secondary global indexes, if any, when querying and, range keys, if any, when scanning. Otherwise you'll get an `Aws::DynamoDB::Errors::ValidationException` either for `Exclusive Start Key must have same size as table's key schema` or `The provided starting key is invalid`
```
richardhsu: Hmm not sure on this wording since confused by "including ... secondary global indexes".

Maybe something like "You need to explicitly define all required keys, including range keys, depending on table or secondary index signatures otherwise you'll get an .."

mnussbaumer:
@richardhsu done! Lmk

pboling:
LGTM 👍

Issue #237, opened 2018-02-16
Strange behavior for rake dynamoid:create_tables


baloran:
I was using `rake dynamoid:create_tables` to launch tests. But an error occured on require of the model.
```ruby
  task :create_tables => :environment do
    # Load models so Dynamoid will be able to discover tables expected.
    Dir[ File.join(Dynamoid::Config.models_dir, "*.rb") ].sort.each { |file| require "./#{file}" }
    if Dynamoid.included_models.any?
      tables = Dynamoid::Tasks::Database.create_tables
      result = tables[:created].map{ |c| "#{c} created" } + tables[:existing].map{ |e| "#{e} already exists" }
      result.sort.each{ |r| puts r }
    else
      puts "Dynamoid models are not loaded, or you have no Dynamoid models."
    end
  end
```

I have changed `require file` to `require "./#{file}"`, and it's working. Ugly, no ?

I have seen this "solution" here: https://stackoverflow.com/a/41991909/2707937

I don't know if is my configuration ? I don't know how to test this.

xoxo

andrykonchin:
I think the correct solution is to change `Dynamoid::Config.models_dir` from default `app/models` to `./app/models`

baloran:
Ok. I made the change !

baloran:
PR made.

baloran:
Ok, I doesn't understant.. sorry @andrykonchin for this pr, I closed.

pboling:
I think the way you have done it, `require "./#{file}"`, despite being ugly, is the correct solution for this use case. 👍

baloran:
I have just changed the models_dir with `./app/models` 🙃

PR #238, opened 2018-02-16
Fix: Config default models_dir

baloran:
Change the config for `:models_dir`.

Refer to #237


andrykonchin:
Actually it makes sense to change default value. Current default value doesn't work well with Rails...

baloran:
I merge ?

Issue #218, opened 2018-01-10
Conditional Updates + Updates without find


pratik60:
**Updating item without find**

I'm looking to implement something along the lines of https://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#update_item-instance_method

The Active Record way of doing this is `Book.update(id, {author: "Carla"})`. This would ideally issue a request to aws to update the item, and avoids the find. 

I am able to do something like this `Book.find(id).update_attributes({author: "Carla"})`, but this probably does a fetch and then issues the update. 

**Conditional Updates**

If we'd access to the DynamoDb client, I would do something like this for a conditional update. 

```
person = Person.find(id)
new_timestamp = Time.now
self.class.dynamodb_client.update_item(
        table_name: self.class.table_name,
        key: { id: id },
        update_expression: "SET last_contacted_at = :new_timestamp",
        condition_expression: "attribute_not_exists(last_contacted_at) OR last_contacted_at = :original_value",
        expression_attribute_values: {
          ":new_timestamp" => new_timestamp,
          ":original_value" => person.last_contacted_at,
        }
      )
```
Again, I couldn't find an equivalent in the Readme/wiki or issues. The closest being here: https://github.com/Dynamoid/Dynamoid/blob/23362c99e357432210782afb2fa475f1491a7230/spec/dynamoid/persistence_spec.rb#L545-L548

How would you approach this?

andrykonchin:
> The Active Record way of doing this is Book.update(id, {author: "Carla"}). This would ideally issue a request to aws to update the item, and avoids the find.

Actually ActiveRecord loads record, sets new attributes values, validates, runs callbacks and saves. It makes sense to have this hight level functionality here in Dynamoid. I am not sure it makes sense to have low level operation what you propose on the Document level.

> If we'd access to the DynamoDb client, I would do something like this for a conditional update

If you need just to access DynamoDB client connected to the server just use `Dynamoid.adapter.client`.

There is a tiny wrapper class what makes things a little simpler `Dynamoid.adapter` and you can make conditional upsert like you want with `Dynamoid.adapter.update_item` (https://github.com/Dynamoid/Dynamoid/blob/master/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb#L385).

andrykonchin:
@pboling 

There is `#upsert` method in `Mongoid` (https://docs.mongodb.com/mongoid/master/tutorials/mongoid-persistence/).

[...]

Issue #119, opened 2016-12-13
How can I enforce an UpdateItem call with changed  attrs instead of a PutItem?


NielsKSchjoedt:
I use a custom (and predictable) hash_key for my items. 
Dynamodb has the possibility of either updating or creating an object, if you use UpdateItem instead of PutItem (a bit like postgres's `upsert`). 

I would like to utilise this for new items. This way I can avoid a roundtrip checking if the object with the hash_key already exists, before calling save, instead It will just be updated with whatever attributes I send. I cannot find any way to do this though? Is there a way using the existing DSL or would that require me to create the functionality from scratch?

pboling:
I do not know if this is implemented in Dynamo.  I don't, and wouldn't, use it for the use case I have with Dynamoid.  If you can't find any reference to `UpdateItem` in the code, then it isn't implemented, and will require someone with vested interest to do it.

andrykonchin:
`.upsert` method was implemented

PR #239, opened 2018-02-24
Add methods for low-level updating

andrykonchin:
* Added `Document.update(id, [range key], attributes)`
* Added `Document.update_fields(id, [range key], attributes, [conditions])`
* Added `Document.upsert(id, [range key], attributes, [conditions])`

`conditions` means `if: { version: 3 }`

`Document.update` loads document, changes attributes and saves. It runs all callbacks and validations.

`Document.update_fields` sends to DynamoDB only changed attributes. It doesn't run callbacks and validations.

`Document.upsert` does the same but it creates new document if there is no document with specified `id`.

All of them return updated document if updating is successful. They use low-level `UpdateItem` api call (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html)

----

Related issues https://github.com/Dynamoid/Dynamoid/issues/218, https://github.com/Dynamoid/Dynamoid/issues/119

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ end
 
 ### Note on set type
 
-There is `if` option to declare the type of set elements. You can use
+There is `of` option to declare the type of set elements. You can use
 `:integer` value only
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -188,6 +188,20 @@ class Document
 end
 ```
 
+### Note on set type
+
+There is `if` option to declare the type of set elements. You can use
+`:integer` value only
+
+```ruby
+class Document
+  include DynamoId::Document
+
+  field :tags, :set, of: :integer
+end
+```
+
+
 #### Magic Columns
 
 You get magic columns of id (string), created_at (datetime), and updated_at (datetime) for free.
--- a/README.md
+++ b/README.md
@@ -523,6 +523,41 @@ User.where("created_at.lt" => DateTime.now - 1.day).all
 
 It also supports .gte and .lte. Turning those into symbols and allowing a Rails SQL-style string syntax is in the works. You can only have one range argument per query, because of DynamoDB's inherent limitations, so use it sensibly!
 
+
+### Updating
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -188,6 +188,20 @@ class Document
 end
 ```
 
+### Note on set type
+
+There is `if` option to declare the type of set elements. You can use
```
pboling: I think you meant 
```
`of` option
```
instead of
```
`if` option
```

lib/dynamoid/document.rb
```diff
@@ -119,6 +119,77 @@ def exists?(id_or_conditions = {})
...
+        options[:conditions] = conditions
+
+        begin
+          new_attrs = Dynamoid.adapter.update_item(table_name, hash_key_value, options) do |t|
+            attrs.symbolize_keys.each do |k, v|
+              t.set k => dump_field(v, attributes[k])
+            end
+          end
+          new(new_attrs)
+        rescue Dynamoid::Errors::ConditionalCheckFailedException
```
pboling: I don't understand this silent rescue with no action taken.  How does this failure get surfaced then?

lib/dynamoid/document.rb
```diff
@@ -119,6 +119,77 @@ def exists?(id_or_conditions = {})
...
```
pboling: I don't understand this silent rescue with no action taken. How does this failure get surfaced then?

lib/dynamoid/document.rb
andrykonchin: @pboling method returns `nil` instead of object. And we can add methods with `!` what will raise some meaningful exceptions. Alternative to model/`nil` is to return `true`/`false` but I like to load attributes and instantiate model for free (AWS doesn't charge this reading operation)

pboling:
LGTM once the readme typo is fixed.  @richardhsu ?

PR #242, opened 2018-03-07
Fix update attributes typo

yoonwaiyan:
`update_attributes` was previously documented as `update_attrubutes`, assuming that it's not a valid word and not the exact method as written in the codebase. Fixed the typo to avoid misguiding.


andrykonchin:
Thank you for the fix

Issue #241, opened 2018-03-06
where.not or ne Operator


yoonwaiyan:
I'm still new to this repo. I've tried to search through the source code and seems like there's no available query similar to `.where.not` in Rails. In AWS DynamoDB console table filters the not equal is available and I checked it's available in the AWS documentation too. Can we have this feature in Dynamoid so that I can make the backend switch from ActiveRecord simpler?

pboling:
You are correct that we haven't formalized a shorthand chained method for the `ne` operator.  @andrykonchin I'm thinking it may no be too hard, but also there may be some unexpected behaviors with chaining multiple conditions, AND vs. OR?  I haven't really used the chaining scoping aspect of the gem yet.

andrykonchin:
@pboling It isn't hard to add `not` chain method except I don't see any way to negate `in` operator. So e.g.
```ruby
Document.where.not('version.in' => [1, 2, 3])
```

will not work. May be I missed something in documentation (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Condition.html).

New condition syntax allows more complex expressions (https://docs.amazonaws.cn/en_us/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html) but we still use deprecated one.

Actually even `ne` operator isn't implemented in `where` conditions. So at least we can add it.

pboling:
OK, let's start with adding support for `ne` operator.  Perhaps after that our goal should be to move to the supported condition syntax?  And *then* add a chained `not`?

andrykonchin:
@pboling Yep, definitely. It will improve querying drastically and make it more SQLish

I think it's right time to consider new minor release. There are two issues I would like to complete before it - this one with `ne` operator and handling not processed items in batch operations.

Does it make sense?

pboling:
Yes!  Looking forward to seeing what you do.  Sorry I don't have more time to help.  I am up to my ears in projects.

PR #243, opened 2018-03-10
Support `NE` condition operator

andrykonchin:
Now we can use `ne` operator in where conditions:

```ruby
Document.where('age.ne' => 10).all
```

Be aware it doesn't work for sort key in QUERY operation (when there is some condition on partition key)

https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Condition.html

Related issue https://github.com/Dynamoid/Dynamoid/issues/241

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -18,6 +18,7 @@ module Dynamoid
       # we declare schema in models
       FIELD_MAP = {
           eq:           'EQ',
+          ne:           'NE',
           gt:           'GT',
           lt:           'LT',
           gte:          'GE',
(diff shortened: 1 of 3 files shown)


pboling:
@andrykonchin We need to add the synopsis of all these changes into the CHANGELOG!

mrkamel:
can this query be supported by indices?
i have a global secondary index on `version`, but dynamoid complains: 

```ruby
pry> Recommendation.where('version.ne': '8a76c71b52b928ed12f5268ed8291d0b').first
Queries without an index are forced to use scan and are generally much slower than indexed queries!
You can index this query by adding index declaration to recommendation.rb:
* global_secondary_index hash_key: 'some-name', range_key: 'some-another-name'
* local_secondary_index range_key: 'some-name'
Not indexed attributes: :version.ne
```

andrykonchin:
What request would you expect Dynamoid to perform in terms of [Scan](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html)/[Query](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html) operations?

Issue #226, opened 2018-01-23
Handling unprocessed_items and unprocessed_keys


pk-n:
how are unprocessed_items for batch_write and unprocessed_keys for batch_read handled?

pk-n:
I came up with a very simple approach for batch_write

`      def batch_write_item table_name, objects, options = {}
        request_items = []
        options ||= {}
        objects.each do |o|
          request_items << { "put_request" => { item: o } }
        end

        begin
          resp = client.batch_write_item(
            {
              request_items: {
                table_name => request_items,
              },
              return_consumed_capacity: "TOTAL",
              return_item_collection_metrics: "SIZE"
            }.merge!(options)
          )
          unless resp.unprocessed_items.empty?
            sleep(5)
            unprocessed_objects = []
            resp.unprocessed_items[table_name].each do |request_item|
              unprocessed_objects << request_item.put_request.item
            end
            batch_write_item(table_name, unprocessed_objects, options)
          end
        rescue Aws::DynamoDB::Errors::ConditionalCheckFailedException => e
          raise Dynamoid::Errors::ConditionalCheckFailedException, e
        end
      end`

pk-n:
I think I can also do the same check where I make a request(outside the gem) to batch_write_item.

andrykonchin:
From the documentation:

If one or more of the following is true, DynamoDB rejects the entire batch write operation:

1. One or more tables specified in the BatchWriteItem request does not exist.

2. Primary key attributes specified on an item in the request do not match those in the corresponding table's primary key schema.

3. You try to perform multiple operations on the same item in the same BatchWriteItem request. For example, you cannot put and delete the same item in the same BatchWriteItem request.

4. There are more than 25 requests in the batch.

5. Any individual item in a batch exceeds 400 KB.
 
6. The total request size exceeds 16 MB.

https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html

@pk-n Your approach solves the problem related to table provision exceeding but doesn't solve listed issues above.

I would consider raising an exception with unprocessed items

pk-n:
Thank you @andrykonchin for replying to my question.
It makes sense. Now I am handling all the six points out side(in my code) the gem where I make a call to batch operation.

pboling:
Accidentally closed.  Should we leave this open for further discussion of ways we can mitigate some of these inside the gem?  I like the idea of raising a custom error that has the unprocessed items embedded.

[...]

PR #246, opened 2018-03-18
Handle unprocessed items in batch operations

andrykonchin:
Exponential backoff is a [recommended](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.BatchOperations) way to handle throughput limits exceeding and throttling on the table.

There are two batch operations - `.find_all` and `.import`.

In this PR we:
* add handling of unprocessed items in batch operations
* allow to use backoff for them

---
Changes:
* `batch_get_items` and `batch_write_item` re-try to handle all unprocessed keys
* `batch_get_items` and `batch_write_item` accept optional block to provide boolean flag if there are unprocessed items in each batch
* add `backoff` and `backoff_strategies` config options
* `find_all` and `import` take into account `backoff` config option

---

There are two simple examples of backoff - constant delay (`:constant`) and [truncated binary exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) (`:exponential`)

Backoff should be specified in following way:

```ruby
Dynamoid.config.backoff = { constant: 2.seconds }
Dynamoid.config.backoff = { exponential: { base_backoff: 0.2.seconds, ceiling: 10 } }
```
User is able to add custom strategies in following way:

```ruby
Dynamoid.config.backoff_strategies[:custom] = lambda do |n|
  -> { sleep  rand(n) }
end
Dynamoid.config.backoff = { custom: 10 }
```
---
Related issue https://github.com/Dynamoid/Dynamoid/issues/226

### DIFF
--- a/README.md
+++ b/README.md
@@ -680,6 +680,8 @@ Listed below are all configuration options.
   Acceptable values - `utc`, `local` (to use system time zone) and time zone name e.g. `Eastern Time (US & Canada)`. Default is `local`
 * `store_datetime_as_string` - if `true` then Dynamoid stores :datetime fields in ISO 8601 string format. Default is `false`
 * `store_date_as_string` - if `true` then Dynamoid stores :date fields in ISO 8601 string format. Default is `false`
+* `backoff` - is a hash: key is a backoff strategy (symbol), value is parameters for the strategy. Is used in batch operations. Default id `nil`
+* `backoff_strategies`: is a hash and contains all available strategies. Default is { constant: ..., exponential: ...}
 
 
 ## Concurrency
@@ -699,6 +701,44 @@ end
 In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleObjectError` if a concurrent process loaded, edited, and saved the same row. Your code should trap this exception, reload the row (so that it will pick up the newest values), and try the save again.
 
 Calls to `update` and `update!` also increment the `lock_version`, however they do not check the existing value. This guarantees that a update operation will raise an exception in a concurrent save operation, however a save operation will never cause an update to fail. Thus, `update` is useful & safe only for doing atomic operations (e.g. increment a value, add/remove from a set, etc), but should not be used in a read-modify-write pattern.
+
+
+### Backoff strategies
+
+
+You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
+
+By some reasons, for instanse table throughput limit is exceeded, some items could be not processed.
+By default operations will be re-run to process this items.
+
+Recommended way to handle throughput limits exceeding and throttling on the table is to use exponential backoff.
+
+There are two default backoff strategies - constant delay and truncated binary exponential backoff.
+By default no backof is used but you can specify one of the default ones:
+
+```ruby
+Dynamoid.configure do |config|
+  config.backoff = { constant: 2.second }
+end
+
+Dynamoid.configure do |config|
+  config.backoff = { exponential: { base_backoff: 0.2.seconds, ceiling: 10 } }
+end
+
+```
+
+You can use your own strategy in following way:
+
+```ruby
+Dynamoid.configure do |config|
+  config.backoff_strategies[:custom] = lambda do |n|
+    -> { sleep rand(n) }
+  end
+
+  config.backoff = { custom: 10 }
+end
+```
+
 
 ## Rake Tasks
 
--- a/README.md
+++ b/README.md
@@ -706,15 +706,15 @@ Calls to `update` and `update!` also increment the `lock_version`, however they 
 ### Backoff strategies
 
 
-You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
+You can use several methods that run efficiently in batch mode like `.find_all` and `.import`.
 
 The backoff strategy will be used when, for any reason, some items could not be processed as part of a batch mode command.
-Operations will be re-run to process this items.
+Operations will be re-run to process these items.
 
 Exponential backoff is the recommended way to handle throughput limits exceeding and throttling on the table.
 
 There are two built-in strategies - constant delay and truncated binary exponential backoff.
-By default no backoff is used but you can specify one of the build-in ones:
+By default no backoff is used but you can specify one of the built-in ones:
 
 ```ruby
 Dynamoid.configure do |config|
@@ -725,6 +725,14 @@ Dynamoid.configure do |config|
   config.backoff = { exponential: { base_backoff: 0.2.seconds, ceiling: 10 } }
 end
 
+```
+
+You can just specify strategy without any arguments to use default presets:
+
+```ruby
+Dynamoid.configure do |config|
+  config.backoff = :constant
+end
 ```
 
 You can use your own strategy in following way:
--- a/README.md
+++ b/README.md
@@ -708,13 +708,13 @@ Calls to `update` and `update!` also increment the `lock_version`, however they 
 
 You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
 
-By some reasons, for instanse table throughput limit is exceeded, some items could be not processed.
-By default operations will be re-run to process this items.
+The backoff strategy will be used when, for any reason, some items could not be processed as part of a batch mode command.
+Operations will be re-run to process this items.
 
-Recommended way to handle throughput limits exceeding and throttling on the table is to use exponential backoff.
+Exponential backoff is the recommended way to handle throughput limits exceeding and throttling on the table.
 
-There are two default backoff strategies - constant delay and truncated binary exponential backoff.
-By default no backof is used but you can specify one of the default ones:
+There are two built-in strategies - constant delay and truncated binary exponential backoff.
+By default no backoff is used but you can specify one of the build-in ones:
 
 ```ruby
 Dynamoid.configure do |config|
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -112,8 +112,7 @@ module Dynamoid
 
         begin
           while items.present? do
-            batch = items.take(BATCH_WRITE_ITEM_REQUESTS_LIMIT)
-            items.shift(BATCH_WRITE_ITEM_REQUESTS_LIMIT)
+            batch = items.shift(BATCH_WRITE_ITEM_REQUESTS_LIMIT)
             requests = batch.map { |item| { put_request: { item: item } } }
 
             response = client.batch_write_item(
@@ -186,8 +185,7 @@ module Dynamoid
           rng = tbl.range_key.to_s
 
           while ids.present? do
-            batch = ids.take(Dynamoid::Config.batch_size)
-            ids.shift(Dynamoid::Config.batch_size)
+            batch = ids.shift(Dynamoid::Config.batch_size)
 
             request_items = Hash.new{|h, k| h[k] = []}
 
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
@@ -84,11 +84,25 @@ module Dynamoid
         @client
       end
 
-      # Puts or deletes multiple items in one or more tables
+      # Puts multiple items in one tables
+      #
+      # If optional block is passed method calls passed block for each written batch of items.
+      # Block receives boolean flag what indicates there are some unprocessed items.
+      #
+      # @example Saves several items to the table testtable
+      #   Dynamoid::AdapterPlugin::AwsSdkV2.batch_write_item('table1', [{ id: '1', name: 'a' }, { id: '2', name: 'b'}])
+      #
+      # @example Pass block
+      #   Dynamoid::AdapterPlugin::AwsSdkV2.batch_write_item('table1', items) do |bool|
+      #     if bool
+      #       puts 'there are unprocessed items'
+      #     end
+      #   end
       #
       # @param [String] table_name the name of the table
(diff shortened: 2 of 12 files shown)


pboling:
@andrykonchin you can ignore the license check failure for now.  I need to configure it more.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
 
 Calls to `update` and `update!` also increment the `lock_version`, however they do not check the existing value. This guarantees that a update operation will raise an exception in a concurrent save operation, however a save operation will never cause an update to fail. Thus, `update` is useful & safe only for doing atomic operations (e.g. increment a value, add/remove from a set, etc), but should not be used in a read-modify-write pattern.
 
+
+### Backoff strategies
+
+
+You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
```
pboling: This sentence is not clear to me.  Can you clarify the meaning abit around "what do efficiently"?

README.md
```diff
@@ -680,6 +680,8 @@ Listed below are all configuration options.
   Acceptable values - `utc`, `local` (to use system time zone) and time zone name e.g. `Eastern Time (US & Canada)`. Default is `local`
 * `store_datetime_as_string` - if `true` then Dynamoid stores :datetime fields in ISO 8601 string format. Default is `false`
 * `store_date_as_string` - if `true` then Dynamoid stores :date fields in ISO 8601 string format. Default is `false`
+* `backoff` - is a hash: key is a backoff strategy (symbol), value is parameters for the strategy. Is used in batch operations. Default id `nil`
+* `backoff_strategies`: is a hash and contains all available strategies. Default is { constant: ..., exponential: ...}
```
pboling: Is this part of config so people can write and plugin their own backoff strategy and add it here?

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
 
 Calls to `update` and `update!` also increment the `lock_version`, however they do not check the existing value. This guarantees that a update operation will raise an exception in a concurrent save operation, however a save operation will never cause an update to fail. Thus, `update` is useful & safe only for doing atomic operations (e.g. increment a value, add/remove from a set, etc), but should not be used in a read-modify-write pattern.
 
+
+### Backoff strategies
+
+
+You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
+
+By some reasons, for instanse table throughput limit is exceeded, some items could be not processed.
```
pboling: Change to something like:

> The backoff strategy will be used when, for any reason, some items could not be processed as part of a batch mode command.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
+By default operations will be re-run to process this items.
```
pboling: This is not the default, unless you turn on exponential backoff, right?  It is turned off by default?

Change to:
> When exponential backoff is configured operations will be rerun to process items that failed.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
+
+Recommended way to handle throughput limits exceeding and throttling on the table is to use exponential backoff.
```
pboling: Change to:

> Exponential backoff is the recommended way to handle throughput limits exceeding and throttling on the table.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
+
+There are two default backoff strategies - constant delay and truncated binary exponential backoff.
```
pboling: Default is a loaded term, let's use built-in instead, since the default is actually `nil`.

> There are two default built-in strategies - constant delay and truncated binary exponential backoff.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
+By default no backof is used but you can specify one of the default ones:
```
pboling: Change to:

> The default configuration is to not use backoff, but instead to *raise an error* when items fail to get processed.
> If you want to enable backoff, customize one of the following to meet your needs:

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -84,36 +84,55 @@ def client
         @client
       end
 
-      # Puts or deletes multiple items in one or more tables
+      # Puts multiple items in one tables
```
pboling: This is not used for deletes?

`table` should be singular as you have it here.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -84,36 +84,55 @@ def client
...
+      #
+      # If optional block is passed method calls passed block for each written batch of items.
```
pboling: Change to:
>  If optional block is passed it will be called for each written batch of items, meaning once per batch.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -84,36 +84,55 @@ def client
...
+      # Block receives boolean flag what indicates there are some unprocessed items.
```
pboling: Change to:
> Block receives boolean flag which is true if there are some unprocessed items, otherwise false.

README.md
andrykonchin: > Is this part of config so people can write and plugin their own backoff strategy and add it here?

Yep. I can imagine people would like to use another implementation of default `exponential` backoff, e.g. not binary and I wouldn't like to write universal exponential strategy.

README.md
andrykonchin: > This is not the default, unless you turn on exponential backoff, right?

@pboling No. I changed default behaviour. Even if backoff isn't set we will process all the items but without any delay. Currently we just lose data at batch writing and reading so I would consider it as a bug fix as well.

pboling: I am confused by this changeset.  It looks like the default for exponential backoff is `nil`, and thus it is not used by default.  Is that correct?

andrykonchin: @pboling As you wrote below `by default` means `built-in`. Will replace it everywhere.

README.md
andrykonchin: @pboling They use batch operations like `batch_get_item` and `batch_write_item`. So they make fewer HTTP requests.

README.md
andrykonchin: @pboling I implemented what was proposed in the issue (https://github.com/Dynamoid/Dynamoid/issues/226#issuecomment-369736839)

> I would like both batch_get_item and batch_write_item to handle unprocessed items themselves and try to read/write unprocessed items until the end. It's a naive approach but it works and it hides problem from a client.

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
andrykonchin: > This is not used for deletes?

andrykonchin:
@pboling Done

README.md
richardhsu: Maybe like "You can use several methods that run efficiently in batch mode like `.find_all`..."

It's unclear to me as well.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
 Calls to `update` and `update!` also increment the `lock_version`, however they do not check the existing value. This guarantees that a update operation will raise an exception in a concurrent save operation, however a save operation will never cause an update to fail. Thus, `update` is useful & safe only for doing atomic operations (e.g. increment a value, add/remove from a set, etc), but should not be used in a read-modify-write pattern.
 
+
+### Backoff strategies
+
+
+You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
+
+The backoff strategy will be used when, for any reason, some items could not be processed as part of a batch mode command.
+Operations will be re-run to process this items.
```
richardhsu: "to process **these** items"

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
+
+Exponential backoff is the recommended way to handle throughput limits exceeding and throttling on the table.
+
+There are two built-in strategies - constant delay and truncated binary exponential backoff.
+By default no backoff is used but you can specify one of the build-in ones:
```
richardhsu: "one of the **built**-in ones"

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -84,36 +84,55 @@ def client
...
-          requests << os.map { |o| { put_request: { item: sanitize_item(o) } } }
-        end
+        items = objects.map { |o| sanitize_item(o) }
 
         begin
-          requests.each do |request_items|
-            client.batch_write_item(
+          while items.present? do
+            batch = items.take(BATCH_WRITE_ITEM_REQUESTS_LIMIT)
+            items.shift(BATCH_WRITE_ITEM_REQUESTS_LIMIT)
```
richardhsu: `shift` should return the batch as well so could just combine the two lines
`batch = items.shift(BATCH_WRITE_ITEM_REQUESTS_LIMIT)`

lib/dynamoid/adapter_plugin/aws_sdk_v2.rb
```diff
@@ -141,19 +180,23 @@ def batch_get_item(table_ids, options = {})
...
           next if ids.blank?
+          ids = Array(ids).dup
           tbl = describe_table(t)
           hk  = tbl.hash_key.to_s
           rng = tbl.range_key.to_s
 
-          Array(ids).each_slice(Dynamoid::Config.batch_size) do |ids|
+          while ids.present? do
+            batch = ids.take(Dynamoid::Config.batch_size)
+            ids.shift(Dynamoid::Config.batch_size)
```
richardhsu: Same can combine `take` and `shift` together

lib/dynamoid/config/backoff_strategies/exponential_backoff.rb
```diff
@@ -0,0 +1,24 @@
...
+      # See https://en.wikipedia.org/wiki/Exponential_backoff
+      class ExponentialBackoff
+        def self.call(opts)
+          base_backoff = opts[:base_backoff]
+          ceiling = opts[:ceiling]
+
+          times = 1
+
+          lambda do
+            power = times <= ceiling ? times - 1 : ceiling - 1
```
richardhsu: Maybe minimum?
`power = [times - 1, ceiling - 1].min` or could even do `power = [times, ceiling].min - 1`

lib/dynamoid/config/backoff_strategies/exponential_backoff.rb
```diff
@@ -0,0 +1,24 @@
+module Dynamoid
+  module Config
+    module BackoffStrategies
+      # Truncated binary exponential backoff algorithm
+      # See https://en.wikipedia.org/wiki/Exponential_backoff
+      class ExponentialBackoff
+        def self.call(opts)
+          base_backoff = opts[:base_backoff]
+          ceiling = opts[:ceiling]
```
richardhsu: Would we want potential defaults or this should always be configured by the user? Feel like providing some base backoff configs so can just out of box say "let's have exponential backoff" without worrying about underlying details is nice.

README.md
```diff
@@ -700,6 +702,44 @@ In this example, all saves to `MyTable` will raise an `Dynamoid::Errors::StaleOb
...
+
+### Backoff strategies
+
+
+You can use several methods what do efficiently in batch mode like `.find_all` and `.import`.
+
+The backoff strategy will be used when, for any reason, some items could not be processed as part of a batch mode command.
+Operations will be re-run to process this items.
+
+Exponential backoff is the recommended way to handle throughput limits exceeding and throttling on the table.
```
richardhsu: Can we also clarify that even AWS "strongly recommends that you use an exponential backoff algorithm", almost makes it feel like we should make it default rather than nil, but I can see that being too strong for some users who may just want it to retry as soon as possible.

README.md
pboling: I think it is a translation error.  If you can try restating it with other words I might be able to understand it.

Perhaps:

> There are some built-in exponential backoff strategy configurations that allow methods like `.find_all` and `.import` to process all items, automatically handling API batch limitations, and reprocessing failed items until they pass.

However, now that I have stated it like that... is it possible that this exponential backoff could encounter an item that will always fail to process, and thus the backoff will be in a retry loop forever?  Does the backoff have an escape hatch after a certain number of tries?

README.md
andrykonchin: @richardhsu 

> but I can see that being too strong for some users

Agree

> Can we also clarify that even AWS "strongly recommends

I'm not good at writing documentation. Feel free to reword/update this section.

andrykonchin:
@richardhsu Done

README.md
richardhsu: We can change it to say:

> Exponential backoff is highly recommended by AWS for retrying and backoff strategies, especially when handling throughput limits being exceeded or throttled on the table.

@pboling does that seem clear to you too?

andrykonchin:
I will check these functionality on real AWS later. Then I will try to fix  the https://github.com/Dynamoid/Dynamoid/issues/251 issue.

@pboling @richardhsu 
Does it make sense to release new minor version after this?

pboling:
Let's get all the changes in the changelog, and then see if there is anything breaking, so we can determine minor vs. major.

Issue #251, opened 2018-03-27
Entry with datetime


arkes:
Hey,

I'm having several entries that I'm unable to delete.
When I'm fetching this entry and reload it, it come back with `Dynamoid::Errors::RecordNotFound: Couldn't find MyModel with 'id'=8fbfc5ee-...`. I'm using an index with a  datetime range.
It seems that the method `dump_field` and `undump_field` are using 2 different time class. `dump_field` use `Time` and `undump_field` use `DateTime`

I'm using 
ruby: `ruby 2.3.3p222`
gem: `dynamoid (2.0.0)`
I tried with other versions of Ruby and Rails.

```
> 10.times { x = Time.current; puts "#{x.to_f} - #{x.to_datetime.to_f} - #{x.to_f == x.to_datetime.to_f}" }
1522135554.9757178 - 1522135554.975718 - false
1522135554.975918 - 1522135554.975918 - true
1522135554.975996 - 1522135554.975996 - true
1522135554.9760509 - 1522135554.976051 - false
1522135554.976595 - 1522135554.976595 - true
1522135554.9766479 - 1522135554.976648 - false
1522135554.9766998 - 1522135554.9767 - false
1522135554.976751 - 1522135554.976751 - true
1522135554.9768 - 1522135554.9768 - true
1522135554.9768488 - 1522135554.976849 - false

> 10.times { x = DateTime.current; puts "#{x.to_f} - #{x.to_time.to_f} - #{x.to_f == x.to_time.to_f}" }
1522135556.917298 - 1522135556.9172978 - false
1522135556.917488 - 1522135556.917488 - true
1522135556.9176 - 1522135556.9176 - true
1522135556.9177 - 1522135556.9177 - true
1522135556.917799 - 1522135556.917799 - true
1522135556.917896 - 1522135556.917896 - true
1522135556.917992 - 1522135556.9179919 - false
1522135556.91809 - 1522135556.9180899 - false
1522135556.919354 - 1522135556.9193542 - false
1522135556.919487 - 1522135556.919487 - true
```

I think the solution should be to just to use the same time class inside `dump_field` and `undump_field`. This issue is still present on master.
I'm happy to create a PR if you want to.

arkes:
I made all the changes to use only `Time`, but the issue is still there (with less occurrences). 
This issue affect reload and destroy and probably update.
[...]

PR #252, opened 2018-04-05
Avoid DateTime to Time converting with losing nanoseconds

andrykonchin:
Fix issue described in https://github.com/Dynamoid/Dynamoid/issues/251


pboling:
@andrykonchin Thanks!  I've been on vacation!

andrykonchin:
@pboling This fix may introduce some incompatibility.

It fixes precision of storing DateTime as float number. All finders don't use type casting mechanism. So if user needs to filter by DateTime field he must cast value manually to Float like
```ruby
Document.find_by_id('id', range_key: activated_at.to_f)
```

As far as `to_f` looses precision it may not work (for finders) if you need to filter by equality.

So from the one hand the fix may cause errors in some situations. From the other hand it fixes usage DateTime as range key what is broken now.

So I would like to release it now and add type casting for finders in the next major release.

pboling:
We need to document the changes that are already merged to master in the changelog so we can figure out the scope of the current changes, and the effect they will need to have on the version number increment, following SemVer 2.0.  I suspect we may already be at a minor or major version bump with the existing changes in master.  So you have bandwidth to start documenting the changes in the changelog?

andrykonchin:
@pboling I've already updated the change log, except for this fix

pboling:
Awesome!  Let's update the changelog for this, and release 2.0.1, assuming we call this a "fix" and not a breaking change.

andrykonchin:
@pboling Done.

Don't you mind adding me to the contributors list in the Readme and the authors list in the gemspec files?

pboling:
@andrykonchin Please add yourself to both! 💯

Issue #254, opened 2018-04-24
Support for rails 5.2?


ryo19910801:
Is this support going to be added?

"activemodel/lib/active_model/attribute_mutation_tracker.rb" added since 5.2
I use in urgent response because it is not supported

override the method in which the error occurred

https://github.com/ryo19910801/Dynamoid/commit/af5e331a742bedd0c05aca7e2780722b96cf1ae1

andrykonchin:
How can we reproduce the issue? Could you please add a code snippet do demonstrate the lack of support?

richardhsu:
I was able to reproduce this in my own by just creating a Dynamoid backed object as well and getting a `nil` exception.

They added `mutations_from_database` as @ryo19910801 mentioned:

```
def mutations_from_database
  unless defined?(@mutations_from_database)
    @mutations_from_database = nil
  end
  @mutations_from_database ||= if @attributes
    ActiveModel::AttributeMutationTracker.new(@attributes)
  else
    NullMutationTracker.instance
  end
end
```

It looks like Rails 5.2 `ActiveModel::Dirty` changed attributes to attribute sets.
Basically if you look at our attributes we return something like:

```
{
  :name => nil,
  :age => nil,
}
```

At creation but seems like ActiveModel now returns:

```
#<ActiveModel::AttributeSet:0x000...
  @attributes=
  { "name" => #<ActiveModel::Attribute::FromDatabase:0x000... @name="name"  ...>
    "age" => #<ActiveModel::Attribute::FromDatabase:0x000... @name="age" ...> }
>
```

Which based on the commit that affects this:
https://github.com/lugray/rails/blob/c3675f50d2e59b7fc173d7b332860c4b1a24a726/activemodel/lib/active_model/attribute.rb

Has defined `changed?` method on the value of `@attributes["name"].changed?` which is the issue.

For the temporary fix would prefer if we were explicit in the exception catch:

```
      NoMethodError:
        undefined method `changed?' for nil:NilClass
```

So then to catch and be specific which even then is still scary but only error we get:
```
rescue NoMethodError => e
  return false if e.message == 'undefined method `changed?' for nil:NilClass'
  raise e
end
```

That way we only raise for this particular mishandling. As for long term in terms of handling this not sure what we'd want to do if we should support similarly. Will continue to experiment.

richardhsu:
[...]

PR #256, opened 2018-04-25
Support Rails 5.2

andrykonchin:
Related issue https://github.com/Dynamoid/Dynamoid/issues/254


lib/dynamoid/dirty.rb
```diff
@@ -5,7 +5,7 @@ module Dirty
 
     module ClassMethods
       def from_database(*)
-        super.tap { |d| d.changed_attributes.clear }
+        super.tap { |d| d.send(:clear_changes_information) }
```
richardhsu: Probably need to have it handle both versions since older versions of rails I don't think will support this so maybe:

```
d.respond_to?(:clear_changes_information) ? d.clear_changes_information : d.changed_attributes.clear
```

And similar to below to handle both

andrykonchin: @richardhsu fixed

andrykonchin:
@richardhsu I would consider independent realization of changes tracking. Depending on Rails internals makes me sad.

richardhsu:
@andrykonchin Yeah makes sense, and yeah their internals are changing so much now.

I've run a few tests with this update and looking good so far. Going to run full suite tomorrow and make sure those pass.

nijikon:
I'm running this in production, so far no issues.

andrykonchin:
@pboling I would like to release new version 2.1.1 with this bug fix only. Does it make sense?

pboling:
@andrykonchin do it!

pboling:
As for depending on Rails internals, I think making this gem a standalone ORM that can be used with other frameworks would be great.  Might as well make that a target goal of a future release.

PR #257, opened 2018-04-27
Add: version to gemspec

walkersumida:
Rspec fails when activemodel version 5.2.
So until support activemodel version 5.2, I think it's better to add `<= 5.1`.


walkersumida:
oh..., sorry.
https://github.com/Dynamoid/Dynamoid/pull/256

PR #261, opened 2018-05-30
Fix README

walkersumida:


andrykonchin:
Thank you for the fix
👍

walkersumida:
@andrykonchin 
Dynamoid is good gem. Thank you 👍

PR #264, opened 2018-06-11
Fix README.md

xbx:
Changed undefined variables to strings.


andrykonchin:
Thank you for the fix

PR #267, opened 2018-06-16
Upgrade AWS SDK to V3

andrykonchin:
So now:
- we use SDK 3
- require only `aws-sdk-core` and `aws-sdk-dynamodb` gems
- client doesn't have to require SDK manually

Instructions https://github.com/aws/aws-sdk-ruby#upgrade-from-version-2


jamtur01:
👍

Issue #37, opened 2016-01-28
Typecasting raises exceptions before validations get to run


sumocoder:
``` ruby
class User
  include Dynamoid::Document
  field :age, :integer

  validates :numericality => {:allow_blank => true, :only_integer => true }
end

# This is raises an exception about "abc" not being an integer
u = User.new(:age => "abc")
```

ActiveRecord attempts to typecast and if it fails it stores nil. It also stores age_before_type_cast. Something similar needs to be done for Dynamoid.

More info here 
http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/BeforeTypeCast.html

andrykonchin:
Fixed

PR #275, opened 2018-07-11
Add type casting

andrykonchin:
Introduce separate mechanisms to:
* dumpt field to DynamoDB type
* undump it to Dynamoid field type
* type cast user value to Dynamoid field type

As a side effect custom types dumping was repaired. See https://github.com/Dynamoid/Dynamoid/pull/275/commits/64c4e2db6e3fbcbd9bf3052d873748d96f348948
The issue was - dumping convertion is used to dump and type cast values.
The goal of the next PRs is to improve type casting and add type casting and dumping to finders.

Related issue https://github.com/Dynamoid/Dynamoid/issues/37


andrykonchin:
@pboling @richardhsu 

Could you please take a look at the PR?

There are two issues to discuss:
* what type casting should be added ([here](https://github.com/Dynamoid/Dynamoid/blob/b84b5fe6453b3c31cecda76970731d0d2a2ad2e3/lib/dynamoid/type_casting.rb))
* do we still need following finders `find_by_composite_key`, `find_all_by_composite_key` and `find_all_by_secondary_index`? Looks like they duplicate functionality of `find/find_by_id` and querying with `where` chain

andrykonchin:
@pboling 
> Also should we deprecate first, for the next release, and then remove for the following release?

Yep. Makes sense. 

TBH I wonder if `find_by_composite_key`, `find_all_by_composite_key` and `find_all_by_secondary_index` are duplicated as far as I didn't use them at all and there may be some unique features I didn't noticed

pboling:
👍 Awesome.

PR #278, opened 2018-07-24
Improve type casting

andrykonchin:
Based on https://github.com/Dynamoid/Dynamoid/pull/275

Changes:
* Add type casting of scalar types like it's done in Rails
* Improve type casting of collection types

Rails type casting:
https://github.com/rails/rails/tree/master/activemodel/test/cases/type
https://github.com/rails/rails/tree/master/activemodel/lib/active_model/type

PR #280, opened 2018-07-26
Type casting in finders

andrykonchin:
Based on https://github.com/Dynamoid/Dynamoid/pull/278

* Type cast sort key passed to finder methods `find`, `find_all` and `find_by_id`
* Fixe some bugs in `.find`
* As a side effect `find_all` and `find_by_id` raise a RecordNotFound errors now.

### DIFF
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -131,10 +131,6 @@ module Dynamoid #:nodoc:
       # @example Update document
       #   Post.update(101, read: true)
       def update(hash_key, range_key_value = nil, attrs)
-        range_key_value = if range_key.present?
-                            Dumping.dump_field(range_key_value, attributes[range_key])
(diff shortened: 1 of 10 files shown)


lib/dynamoid/document.rb
```diff
@@ -115,7 +115,11 @@ def build(attrs = {})
       def exists?(id_or_conditions = {})
         case id_or_conditions
         when Hash then where(id_or_conditions).first.present?
-        else !!find_by_id(id_or_conditions)
+        else
+          begin
+            find_by_id(id_or_conditions)
+          rescue Dynamoid::Errors::RecordNotFound
+          end
```
richardhsu: Return `false` so it returns Boolean.

andrykonchin: @richardhsu Good catch

PR #281, opened 2018-08-03
Deprecate old style finders

andrykonchin:
Following finders are deprecated:
- find_all
- find_by_id
- find_by_composite_key
- find_all_by_composite_key
- find_all_by_secondary_index
- and dynamic finders

All these methods can be replaced with `find` and `where` methods.

The goal is to make API more compatible to ActiveRecord and Mongoid.

The `find_by`/`find_by!` methods are missing only. Will be added in separate PR.

### DIFF
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -117,7 +117,7 @@ module Dynamoid #:nodoc:
         when Hash then where(id_or_conditions).first.present?
         else
           begin
-            find_by_id(id_or_conditions)
+            find(id_or_conditions)
             true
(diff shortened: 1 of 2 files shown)


pboling:
@andrykonchin This looks great.  Once we add the `find_by` I would like to update the deprecation notice for `find_by_id` to point to `find_by(id: id)`.

In Rails there is a noisy difference between `find`, which raises an error when not finding the target, and `find_by` which quietly returns `nil`.  I'd like to continue that expectation in Dynamoid for POLS.

andrykonchin:
@pboling Yep. Makes sense

PR #282, opened 2018-08-03
[WIP] Use string format for date and datetime by default

andrykonchin:
I propose to store Date and DateTime in string format **by default** because
- it's more universal format
- it's more readable
- and easier to work with.


pboling:
@andrykonchin I've merged #283 and think this proposal is reasonable.  Let's move forward with it!

andrykonchin:
I would like to postpone this change and to test time zone related stuff for a while.

Issue #273, opened 2018-07-05
Is there a way to fetch all documents without getting the scan warning?


knovoselic:
Hi,

I have a use-case where I need to fetch all documents from a specific collection. Doing

    TheCollection.all.to_a

issues a scan warning. Is there a way to do this without scanning (probably not, right?), or at least without printing out the scan warning just for this case (when intentionally trying to get all documents)?

I think it doesn't make sense to issue scan warning in this case, as it's obvious that I want to read the full collection. Would it be acceptable to change the code so that the warning doesn't show up if query criteria is empty?

andrykonchin:
Good point.

> Would it be acceptable to change the code so that the warning doesn't show up if query criteria is empty?

Yes, it makes sense.

knovoselic:
Cool. I can work on the pull request in about two weeks, depending on my current work priorities. What's your policy on cutting a new gem version? Do you wait for a bunch of fixes and then push them together? Any idea when the next version is going out?

PR #286, opened 2018-08-14
Disable scan warning when intentionally loading all items from a collection

knovoselic:
This is a fix/improvement discussed in #273. The change modified scan warning not to show if user is intentionally loading full collection into memory (that is, when there is not query/criteria).


andrykonchin:
Thanks

PR #287, opened 2018-08-14
Fix inability to disable logging

ghiculescu:
Before this fix, if you did `config.logger = nil`, it would be overridden and replaced with the default logger if you ever called `config.logger` again afterwards. Thus making it difficult to actually disable the logger.

With this fix, I create a new but useless logger.


andrykonchin:
LGTM 👍 
Thanks

PR #288, opened 2018-08-14
Add store_boolean_as_native global config option

andrykonchin:
The option is similar to the `store_as_native_boolean` field option and means the same.

By default the option is true so any boolean field will be stored as native DynamoDB's boolean value.

### DIFF
--- a/lib/dynamoid/undumping.rb
+++ b/lib/dynamoid/undumping.rb
@@ -156,10 +156,15 @@ module Dynamoid
 
     class BooleanUndumper < Base
       def process(value)
-        if ['t', true].include? value
-          true
-        elsif ['f', false].include? value
-          false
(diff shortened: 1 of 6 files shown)


lib/dynamoid/undumping.rb
```diff
@@ -156,10 +156,15 @@ def process(value)
...
-        elsif ['f', false].include? value
-          false
+        store_as_boolean = if @options[:store_as_native_boolean].nil?
+                             Dynamoid.config.store_boolean_as_native
+                           else
+                             @options[:store_as_native_boolean]
+                           end
+        if store_as_boolean
+          !!value
+        elsif ['t', 'f'].include?(value)
```
pboling: Would like to make this array a constant within the class.  No need to instantiate again on every call.

PR #294, opened 2018-09-08
Fix flaky specs related to numeric datetime format

andrykonchin:
There is an issue with numeric format of `datetime` fields.

Dynamoid stores `datetime` field as a float number and may lose precision while convert it to needed time zone etc e.g stored value `1536429694.118027` may be loaded as `1536429694.118026999`.

So I added warning to not to use `datetime` as a range key, because when precision is lost user cannot load a document with hash key/range key.

Later we can store `datetime` without float part by default - just seconds - and add config option to preserve current behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ end
 
 #### Note on datetime type
 
-By default datetime fields are persisted as UNIX timestamps with millisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_string` to `true`
+By default datetime fields are persisted as UNIX timestamps with millisecond precision in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_string` to `true`
 
 ```ruby
 class Document
--- a/README.md
+++ b/README.md
@@ -165,13 +165,13 @@ By default date fields are persisted as days count since 1 January 1970 like UNI
 class Document
   include DynamoId::Document
 
-  field :sent_at, :datetime, store_as_string: true
+  field :sent_on, :date, store_as_string: true
 end
 ```
 
 #### Note on datetime type
 
-By default datetime fields are persisted as UNIX timestamps with milisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_string` to `true`
+By default datetime fields are persisted as UNIX timestamps with millisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_string` to `true`
 
 ```ruby
 class Document
@@ -180,6 +180,14 @@ class Document
   field :sent_at, :datetime, store_as_string: true
 end
 ```
+
+WARNING: Fiels in numeric format are stored with nanoseconds as a fraction part and precision could be lost.
+That's why `datetime` field in numeric format shouldn't be used as a range key.
+
+There are following options if you need to use `datetime` field as a range key:
+* to use string format
+* to store `datetime` values without milliseconds e.g. cut them
+  manually with `change` method - `Time.now.change(usec: 0)`
 
 #### Note on set type
 
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ class Document
 end
 ```
 
-WARNING: Fiels in numeric format are stored with nanoseconds as a fraction part and precision could be lost.
+WARNING: Fields in numeric format are stored with nanoseconds as a fraction part and precision could be lost.
 That's why `datetime` field in numeric format shouldn't be used as a range key.
 
 There are following options if you need to use `datetime` field as a range key:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -165,13 +165,13 @@ By default date fields are persisted as days count since 1 January 1970 like UNI
...
 
-  field :sent_at, :datetime, store_as_string: true
+  field :sent_on, :date, store_as_string: true
 end
 ```
 
 #### Note on datetime type
 
-By default datetime fields are persisted as UNIX timestamps with milisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_string` to `true`
+By default datetime fields are persisted as UNIX timestamps with millisecond precission in DynamoDB. If you prefer datetimes to be stored as ISO-8601 formatted strings instead then set `store_as_string` to `true`
```
pboling: `precission` should be `precision`

spec/dynamoid/dumping_spec.rb
```diff
@@ -174,19 +174,6 @@
         expect(raw_attributes(obj)[:sent_at]).to eql(BigDecimal('1532131200.0'))
       end
 
-      it 'does not loose precision and can be used as sort key', :bugfix do
```
pboling: Could we keep this spec in another form?  Do we already have specs testing the precision retention of the alternative methods?

spec/dynamoid/persistence_spec.rb
```diff
@@ -681,12 +680,12 @@ def reload_address
...
     it 'uses dumped value of sort key to call UpdateItem' do
       klass = new_class do
-        range :activated_at, :datetime
+        range :activated_on, :date
         field :name
       end
       klass.create_table
 
-      obj = klass.create!(activated_at: Time.now, name: 'Old value')
+      obj = klass.create!(activated_on: Date.today, name: 'Old value')
```
pboling: `Date.today` is dangerous and should generally be avoided as it is not Rails time zone aware.  Especially in specs where we might have people copying the patterns in their own code.  Use `Time.zone.today` instead.

Ref: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/Date

spec/dynamoid/persistence_spec.rb
```diff
@@ -739,13 +738,13 @@ def reload_address
   context 'delete' do
     it 'uses dumped value of sort key to call DeleteItem' do
       klass = new_class do
-        range :activated_at, :datetime
+        range :activated_on, :date
       end
 
-      obj = klass.create!(activated_at: Time.now)
+      obj = klass.create!(activated_on: Date.today)
```
pboling: `Date.today` is dangerous and should generally be avoided as it is not Rails time zone aware.  Especially in specs where we might have people copying the patterns in their own code.  Use `Time.zone.today` instead.

Ref: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/Date

spec/dynamoid/dumping_spec.rb
andrykonchin: @pboling Could you please clarify what form you mean?

spec/dynamoid/persistence_spec.rb
andrykonchin: @pboling Completely agree with you. In Rails application we should use `Time.zone.now` and `Time.zone.today` (and their synonyms).

But it isn't a Rails application. Dynamoid doesn't rely on `Rails.config.timezone`. It could use it 
 (as default value for `application_timezone`) in order to simplify integration with Rails.

IMHO in specs it doesn't matter and it only couples Dynamoid with Rails even more, but Dynamod can be used apart from Rails.

andrykonchin:
@pboling Thank you for the review 👍

spec/dynamoid/persistence_spec.rb
pboling: I'm ok with that reasoning 👍

spec/dynamoid/dumping_spec.rb
pboling: Oh I think I misunderstood.  So there is no current way to not lose precision, but that will be added back in a future PR. 👍

andrykonchin: @pboling 

Yep. The issue is reproduced only with JRuby constantly and isn't reproduced with MRI...

But there is no any way to ensure that time related calculations don't lead to losing precision even on MRI unless core team declares this.

But they say only:
> The lowest digits of to_f and nsec are different because IEEE 754 double is not accurate enough to represent the exact number of nanoseconds since the Epoch.
> The more accurate value is returned by nsec.

https://ruby-doc.org/core-2.5.0/Time.html#nsec-method

andrykonchin:
@pboling Have you completed the review? (The PR still isn't approved).

It's the last PR I would like to merge before the major release so I am waiting to somebody's approving.

pboling:
I am going to test out this new release/master HEAD on my installation as soon as I can!

andrykonchin:
@pboling I've already prepared change log and it's enough I hope. We don't need any upgrading guide.

> I am going to test out this new release/master HEAD on my installation as soon as I can!

Great. Looking forward to your feedback

andrykonchin:
@pboling 

ping

Did you have any chance to run specs in your project agains the `master` branch?

pboling:
Not yet!  Thanks for ping.  I will make an effort to do this today.

Issue #297, opened 2018-09-20
What if an existing DB has a field with a name that's reserved by dynamoid (e.g. "type")


launchbyte-dev:
We're working with a DB that has a table with a field named "type".  

When we try to access data from that table with values in the "type" field (e.g. type="XXX"), we get an error: "NameError: Uninitialized constant XXX.

When we try to access data from that table without values in the "type" field, it returns the record correctly, but, with the table name as the value for the type field.

Is there a way to tell dynamoid to treat a field named "type" specially in some way?  Maybe give it an alias under the covers or something?

Thanks!

pboling:
Here is the list of reserved words in DynamoDB:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html

Please do not use those words to name tables, attributes or other objects.

From the documentation:
> Although DynamoDB allows you to use these reserved words and special characters for names, we recommend that you avoid it because you have to define placeholder variables whenever you use these names in an expression.

So while it is possible to support reserved words as object names, we have no plans to support it.  If you would like to see support you might need to fork the repo, and open a pull request adding that support.

andrykonchin:
The issue is that Dynamoid uses `type` field to support `STI` like it's done on Rails ActiveRecord. So `type` value is treated as a class name of some class hierarchy. 

There is a way to handle this issue In Rails by specifying another column name for STI with [inheritance_column](https://apidock.com/rails/ActiveRecord/Base/inheritance_column/class) in the model.

Looks like it's a good feature to implement in Dynamoid.

@pboling @richardhsu Does it make sense?

pboling:
I would be happy to see it implemented @andrykonchin  - I just don't have time to do it myself. 👍

andrykonchin:
@pboling Great. Will work on it right after releasing 3.0 version.

andrykonchin:
Fixed

PR #304, opened 2018-10-06
Add inheritance field option

andrykonchin:
Related issue https://github.com/Dynamoid/dynamoid/issues/297#issuecomment-423536619

Added new document option `inheritance_field` to allow using another field instead of `type` to support STI.

It's similar to Rails' [inheritance_column](https://apidock.com/rails/ActiveRecord/Base/inheritance_column/class) method

So there is an example:

```ruby
class Car
  include Dynamoid::Document
  table inheritance_field: :my_new_type

  field :my_new_type
end

c = Car.create
c.my_new_type
#=> "Car"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,23 @@ animal.class
 #=>  Cat
 ```
 
+If you already have DynamoDB tables and `type` field already exists and has its own semantic it leads to conflict.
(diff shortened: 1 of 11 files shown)


jonathan-sousa:
❤️

PR #306, opened 2018-10-13
Add NullLogger

andrykonchin:
Use fake logger (`NullLogger`) instead of writing to `/dev/null` if logging is disabled.

I consider using `/dev/null` to be smell and it shouldn't work on Windows.

Related PR https://github.com/Dynamoid/dynamoid/pull/287


pboling:
I prefer using the actual dependency over copying it in here.  Dependencies, even simple ones, allow for the specs to be managed outside this project.  If that library needs updates with new versions of Ruby, we do not want that maintenance burden on this project, we just want to update the gem, automated even, with Depfu.

The genius of GNU linux tools is very small programs that do a single thing well, and are independently maintained, but work well together.  Having small, single class, gems is not a bad thing.

I support adding new dependencies to this gem.  It also reduces pointless memory bloat for the NullLogger class if a project is already using it, then it can be reused, rather than defined all over again.

andrykonchin:
@pboling Done

PR #310, opened 2018-10-26
Fix list_tables method to return names of all tables, not just first page

knovoselic:
Old implementation returns only the first page of `list_tables` call. New implementation pages through the results to return all table names.


andrykonchin:
Looks good 👍 . Thanks for the fix

PR #311, opened 2018-10-31
fix consistent_read option of Finders#find

kokuyouwind:
In [README.md#Consistent Reads](https://github.com/Dynamoid/dynamoid/tree/v2.2.0#consistent-reads), `Model#find` provides `consistent_read` option.

> Address.find(address.id, consistent_read: true)  # Find an address, ensure the read is consistent.

But it does not work.

I fix this by passing `consistent_read` option to client in `AwsSdkV3#get_item`.


andrykonchin:
👍 
Looks good

PR #312, opened 2018-11-05
Add Ability to specify timeout connection settings

lulu-ulul:
We can use [Aws configuration](https://github.com/Dynamoid/dynamoid#aws-configuration) but it's not specified for Dynamoid.

This change adds ability to specify timeout connection settings specific to Dynamoid.

I referred default values to following URL.
https://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html


andrykonchin:
Yeah looks good.

It makes sense to support more AWS options (see [complete list)](https://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#initialize-instance_method) but it may require adding some kind of namespace/prefix for AWS specific options

PR #324, opened 2019-01-10
Order of `global_secondary_index` call matters.

cabello:
Hi 👋 

Just improving docs a tiny bit by letting developers know that the index call must come after fields have been defined.

Thanks! 💚

### DIFF
--- a/README.md
+++ b/README.md
@@ -684,7 +684,7 @@ class User
   field :name
   field :age, :number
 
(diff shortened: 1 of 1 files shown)


pboling:
Perhaps it would be even more helpful to have Dynamoid rescue whatever error happens when the definition is reversed, and print a helpful message instructing the developer to correct the order.

PR #326, opened 2019-01-11
Travis CI. Fix specific bundler version installation

andrykonchin:
Specify `activemodel` instead of `rails` in specs.

There is an issue with running specs against Rails 4.2 because this version requires `bundler < 2.0`. 

Travis CI provides Ruby with preinstalled Bundler and makes it `default` gem (so you cannot uninstall it). After releasing Bundler 2 TravisCI preinstalls it and it breaks our specs.

As far as only Rails limits bundler version but not `activemodel` and Dynamoid depends on `activemodel` only lets get rid of Rails in specs

### DIFF
--- a/gemfiles/rails_4_2.gemfile
+++ b/gemfiles/rails_4_2.gemfile
@@ -6,6 +6,6 @@ source 'https://rubygems.org'
 
 gem 'nokogiri', '~> 1.6.8'
 gem 'pry-byebug', platforms: :ruby
-gem 'rails', '~> 4.2.0'
+gem 'activemodel', '~> 4.2.0'
 
 gemspec path: '../'
--- a/gemfiles/rails_5_0.gemfile
+++ b/gemfiles/rails_5_0.gemfile
@@ -5,6 +5,6 @@
 source 'https://rubygems.org'
 
 gem 'pry-byebug', platforms: :ruby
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -24,9 +24,11 @@ before_install:
...
   - docker ps
 
-after_install:
-  - gem install bundler -v 1.16.2
-  - bundle install
+  # Remove this recommand after dropping Rails 4.2 support
+  # Rails 4.2 requires bundler < 2.0
+  - gem uninstall bundler
+  - gem install bundler -v '< 2'
+  - bundle --version
```
pboling: Cool!

pboling:
> don't require the whole `rails` gem but only `activemodel` (because only rails gem limits bundler version) in specs

This would be a great step forward, as it would also allow people using `activemodel` outside Rails, like in a Sinatra app, to use Dynamoid.  It may require some refactoring away from some rails-isms though.  I think that's generally a positive move.

andrykonchin:
@pboling Actually Dynamoid already depends on `activemodel` only (https://github.com/Dynamoid/dynamoid/blob/master/dynamoid.gemspec#L43)

Rails is required only in test environment on CI so we could drop it right now.

PR #340, opened 2019-04-03
Choose the best global secondary index which consider range key

ryz310:
I use dynamoid in my model as following:

```rb
class User
  include Dynamoid::Document

  table name: :line_users, key: :id

  field :parent_id, :string
  field :some_value, :string
  field :some_date_at, :datetime

  global_secondary_index hash_key: :parent_id, range_key: :some_value, projected_attributes: :all
  global_secondary_index hash_key: :parent_id, range_key: :some_date_at, projected_attributes: :all
end
```

So I would like to search via query like a following code:

```rb
User.where(parent_id: parent_id, 'some_date_at.lte': 1.day.ago)
```

But dynamoid chooses the first finding GSI because it is not consider range key in the GSI selection processing.

https://github.com/Dynamoid/dynamoid/blob/48d167dc06cdb6b9a223ffbc56681f9901c48066/lib/dynamoid/criteria/chain.rb#L315-L321

I modified it to consider the range key.

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -312,16 +312,19 @@ module Dynamoid #:nodoc:
         # Chooses the first GSI found that can be utilized for the query
         # But only do so if projects ALL attributes otherwise we won't
         # get back full data
+        result = false
         source.global_secondary_indexes.each do |_, gsi|
           next unless query.keys.map(&:to_s).include?(gsi.hash_key.to_s) && gsi.projected_attributes == :all
+          next if @range_key.present? && !query_keys.include?(gsi.range_key.to_s)
+
           @hash_key = gsi.hash_key
           @range_key = gsi.range_key
           @index_name = gsi.name
-          return true
+          result = true
         end
 
         # Could not utilize any indices so we'll have to scan
-        false
+        result
       end
 
       # Start key needs to be set up based on the index utilized
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -598,6 +598,7 @@ describe Dynamoid::Criteria::Chain do
           field :gender
 
           global_secondary_index hash_key: :city, range_key: :age, name: :cityage, projected_attributes: :all
+          global_secondary_index hash_key: :city, range_key: :gender, name: :citygender, projected_attributes: :all
           global_secondary_index hash_key: :email, range_key: :age, name: :emailage, projected_attributes: :all
         end
       end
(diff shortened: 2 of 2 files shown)


lib/dynamoid/criteria/chain.rb
```diff
@@ -312,16 +312,19 @@ def key_present?
         # Chooses the first GSI found that can be utilized for the query
         # But only do so if projects ALL attributes otherwise we won't
         # get back full data
+        result = false
         source.global_secondary_indexes.each do |_, gsi|
           next unless query.keys.map(&:to_s).include?(gsi.hash_key.to_s) && gsi.projected_attributes == :all
+          next if @range_key.present? && !query_keys.include?(gsi.range_key.to_s)
```
andrykonchin: Looks like I wanted to check GSI's range key, but not the range key of main table, so it should be

```ruby
next if gsi.range_key.present? &&  ...
```

Does it make sense?

andrykonchin: If my assumption above is correct there are several issues in proposed solution:

1. Result depends on order of GSI declarations in case when we have GSI with range key and GSI without it and  both of them are suitable (both hash keys are involved into query and range key of the first GSI is involved as well). Only the last GSI in the `source.global_secondary_indexes` list will be returned as result, even if it's GSI without range key
2. If there is only one suitable GSI among all the declared in model (so its range key is involved into query) but it has range key not involved into query - the GSI will be skipped as well what isn't expected result.

andrykonchin:
Great idea 👍.
Though I left some comments

ryz310:
Thank you for your review 😄
I am busy now so I will do it a few days later 👍

lib/dynamoid/criteria/chain.rb
ryz310: > Looks like GSI's range key should be checked, but not the range key of the main table. So it should be
> 
> ```ruby
> next if gsi.range_key.present? &&  ...
> ```
> 
> Does it make sense?

That code mean is that it will skip if the range key is already chosen and current GSI's range key is not match query.  So the first time in iteration, this logic behaves like same before modified.
Why do you think GSI's range key should be checked?

ryz310: > 1. Result depends on order of GSI declarations in case when we have GSI with range key and GSI without it and  both of them are suitable (both hash keys are involved into query and range key of the first GSI is involved as well). Only the last GSI in the `source.global_secondary_indexes` list will be returned as result, even if it's GSI without range key

Exactly. But is it troubled?
Please tell me why, for example it makes bad performance etc.

ryz310: > 2\. If there is only one suitable GSI among all the declared in model (so its range key is involved into query) but it has range key not involved into query - the GSI will be skipped as well what isn't expected result.

Is it so? If the assumption is correct, this spec should be failed, isn't it?

https://github.com/Dynamoid/dynamoid/blob/e53f943a39cf62b1f56ac6174e44520e0ce41785/spec/dynamoid/criteria/chain_spec.rb#L641-L646

andrykonchin:
> That code mean is that it will skip if the range key is already chosen and current GSI's range key is not match query. So the first time in iteration, this logic behaves like same before modified.

I see. It looks good 👍

Issue #338, opened 2019-04-01
Non Skip-Take Pagination


arjes:
Hello,

DynamoDB Pagination is a bit different then the skip take (LIMIT & OFFSET driven) pagination we are most developers are used to. Dynamoid offers wrappers to make DynamoDB pagination look like this, but it isn't the native pagination for DynamoDB.

For my use case it is desirable to use the default DynamoDB pagination style, but this isn't exposed anywhere in Dynamoid that I could find. I can add the feature and submit a PR, but I would like to start a discussion on how this should be exposed to the consumers of the library.

So here are my first two ideas:

```ruby
# Given some query...
query = Model.where(foo: 'bar').start(params[:restart_key])

# Return a tuple
records, restart_key = query.next_page

# Return a Hash
data = query.next_page # => {records: [...], restart_key: 'foobar'}
```

pboling:
I really like this idea!

richardhsu:
What would be the difference of this and the limit features? See https://github.com/Dynamoid/dynamoid#limits

The way you'd use it similar to README and should work similar to the following:
```
records = Model.where(foo: 'bar').record_limit(5)
next_5_records = Model.where(foo: 'bar').record_limit(5).start(records.last)

# Or to get them all with your own pagination
bookmark = nil
loop do
  records = Model.where(foo: 'bar').record_limit(5).start(bookmark)
  bookmark = records.last
  process_records(records)
end
```

I believe `start` given a `nil` input will just ignore it.

andrykonchin:
@arjes Could you please clarify what problem you do encounter and would like to solve with your proposal?

The gem isn't documented well so there may be already suitable solution.

arjes:
@andrykonchin In the end I am looking to only retrieve one page s DynamoDB defines it at a time. I've been in and out of the code quite a bit but mostly in V2. We are using DynamoDb fairly heavily at work and have a few [extensions for dynamoid](https://github.com/GetTerminus/dynamoid-advanced-where)

In the end I am trying to solve a problem of efficiency, DynamoDB defines pages essentially by [data size](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html).

> DynamoDB paginates the results from Query operations. With pagination, the Query results are divided into "pages" of data that are 1 MB in size (or less). An application can process the first page of results, then the second page, and so on.

[...]

PR #341, opened 2019-04-06
Add non skip take pagination

arjes:
Partial completion of #338 

This is still in progress, but should have 100% of tests passing. I wanted to put it up to show the approach I have taken incase I'm headed down a path you disagree with.

Basics:
Scan and query now return page at at time. I didn't find documentation on scan and query, so I assume it is an internal method. However if you have been advising users to leverage scan and query directly, this would be a breaking change for them. In this case I can rename them to scan_page etc and maintain existing query/scan behaviors.

Arrays are bubbled up and flattened at the last minute outbound.

At this point adding an each_page, or find_in_pages method is fairly trivial (which will also remove the method calls trailing the `end`).

Most of the changes in this PR are making the specs green with as little change as possible.

I will submit the restart hash generation method as a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ Address.record_limit(10_000).batch(100).each { … } # Batch specified as part o
 The implication of batches is that the underlying requests are done in the batch sizes to make the request and responses
 more manageable. Note that this batching is for `Query` and `Scans` and not `BatchGetItem` commands.
 
-#### DynamoDB Native Pages
+##### DynamoDB Native Pages
 At times it can be useful to rely on DynamoDB [default pages](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Pagination)
 instead of fixed pages sizes. Each page results in a single Query or Scan call
 to DyanmoDB, but return an unknown number of records.
--- a/README.md
+++ b/README.md
@@ -580,6 +580,20 @@ Address.record_limit(10_000).batch(100).each { … } # Batch specified as part o
 The implication of batches is that the underlying requests are done in the batch sizes to make the request and responses
 more manageable. Note that this batching is for `Query` and `Scans` and not `BatchGetItem` commands.
 
+#### Pagination
+At times it can be useful to rely on DynamoDB [default pages](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Pagination)
+instead of fixed pages sizes. Each page results in a single Query or Scan call
+to DyanmoDB, but return an unknown number of records.
+
+Access to the native DynamoDB pages can be gathered via the `find_by_pages`
+method, which yields arrays of records.
+
+```ruby
+Address.find_by_pages do |addresses|
+  # have an array of pages
+end
+```
+
 #### Sort Conditions and Filters
 
 You are able to optimize query with condition for sort key. Following operators are available: `gt`, `lt`, `gte`, `lte`,
--- a/README.md
+++ b/README.md
@@ -584,9 +584,9 @@ more manageable. Note that this batching is for `Query` and `Scans` and not `Bat
 ##### DynamoDB Native Pages
 At times it can be useful to rely on DynamoDB [default pages](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.Pagination)
 instead of fixed pages sizes. Each page results in a single Query or Scan call
-to DyanmoDB, but return an unknown number of records.
+to DyanmoDB, but returns an unknown number of records.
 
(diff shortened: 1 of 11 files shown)


arjes:
Also i just saw the ruby support is for the project is `2.0.0-p648 2.1.10 2.2.6 2.3.3 2.4.1 jruby-9.1.8.0` and `itself` was added in 2.2 i believe. I will refactor away from that.

andrykonchin:
Looks good 👍 

The `query`/`scan` adapter methods are definitely an internal interface and could be changed easily. So it's OK to return pages instead of hashes.

From the other side I like the idea to not modify them and just add new methods (`query_page `/`scan_page `) on the adapter layer. It may lead to fewer changes.

lib/dynamoid/criteria/chain.rb
```diff
@@ -140,13 +140,13 @@ def records
           records_via_query
         else
           records_via_scan
-        end
+        end.lazy.flat_map(&:itself)
```
andrykonchin: `records_via_query `/`records_via_scan` mean "records" not "pages".

So it's reasonable either to make them return lazy collection of hashed (like they did before) or rename them to `pages_via_query`/`pages_via_scan`

arjes: Sounds good, this afternoon is gonna be moved to fixing stuff like that and creating the final paged public method. Yesterday was just getting my hooks into it all and specs green again.

Just didn't want to get too far before showing this since it is a fairly fundamental change in internal datagram

README.md
```diff
@@ -580,6 +581,20 @@ Address.record_limit(10_000).batch(100).each { … } # Batch specified as part o
...
+to DyanmoDB, but returns an unknown number of records.
+
+Access to the native DynamoDB pages can be obtained via the `find_by_pages`
+method, which yields arrays of records.
+
+```ruby
+Address.find_by_pages do |addresses|
+  # have an array of pages
+end
+```
```
arjes: My next PR adding the restart hashes will continue with this section for both fixed limit and native pages.

lib/dynamoid/criteria/chain.rb
```diff
@@ -128,6 +128,10 @@ def each(&block)
         records.each(&block)
       end
 
+      def find_by_pages(&block)
```
arjes: What is the final verdict on the  page for this, if you would prefer pages I can move that method up

lib/dynamoid/criteria/chain.rb
```diff
@@ -136,27 +140,41 @@ def each(&block)
       #
       # @since 0.2.0
       def records
+        pages.lazy.flat_map { |i| i }
```
arjes: the explicit lazy turns out to be needed here, or the flat_map ended up calling scan/query

arjes:
@andrykonchin Taking this off Draft status and going to branch to work on `generate_start_hash` or some such

arjes:
Eh this needs more work. something I did make jruby not like the enums wraped in enums. of course fixing that in jruby angered mri.

arjes:
So there is a difference between MRI and JRuby with yeilding to an enumerator and this is why my change of yielding an array is failing. See these examples below.
```ruby
# jruby
Enumerator.new do |y|
  y << [ 1234 ]
end.to_a
# => [[[1234]]]
```

```ruby
# mri
Enumerator.new do |y|
   y << [ 1234 ]
end.to_a
# => [[1234]]
```

Do you have _any_ idea what is up with this? I'm going to try converting to enum_for and normal yield to see if that fixes it

andrykonchin:
> Also i just saw the ruby support is for the project is 2.0.0-p648 2.1.10 2.2.6 2.3.3 2.4.1 jruby-9.1.8.0 and itself was added in 2.2 i believe. I will refactor away from that.

Actually support of the old versions was ended in v3.0. So `itself` could be used without any limitation

According to Readme.md:

> Dynamoid supports Ruby >= 2.3 and Rails >= 4.2.

lib/dynamoid/criteria/chain.rb
andrykonchin: The name `find_by_pages` looks good.

andrykonchin:
> Do you have any idea what is up with this? I'm going to try converting to enum_for and normal yield to see if that fixes it.

It's definitely an actual bug in JRuby and looks like `enum_for` works. 

Good job 👍

bmalinconico:
So I turns out the bug is not present using `yield`, so  enum_for worked. It sucks to have to create the two signatures, with block and without, but here we are.

In order to have the scan warning issued on creation but before hydration I had to move it around. If you are okay with it only showing once iteration begins, it can be moved after the end _for and the spec updated

PR #342, opened 2019-04-09
Adding metadata to page iteration

arjes:
Adds the metadata for each pages as a second parameter `find_by_pages`.

While I was testing I also found a bug with all of the criteria delegation such as `Model.each`, and the `Model.find_by_pages` method I added. The specs didn't catch it with mine because `match_array` coerces it to an array. The test for criteria `.each` method had a false positive because no `expect` was ever called.

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
@@ -536,7 +536,8 @@ module Dynamoid
         table = describe_table(table_name)
 
         Query.new(client, table, options).call.each do |page|
-          yield page.items.map{ |row| result_item_to_hash(row) }
+          yield page.items.map { |row| result_item_to_hash(row) },
+            { last_evaluated_key: page.last_evaluated_key }
         end
       end
 
@@ -562,10 +563,12 @@ module Dynamoid
       # @todo: Provide support for various options http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#scan-instance_method
       def scan(table_name, conditions = {}, options = {})
         return enum_for(:scan, table_name, conditions, options) unless block_given?
+
         table = describe_table(table_name)
 
         Scan.new(client, table, conditions, options).call.each do |page|
-          yield page.items.map{ |row| result_item_to_hash(row) }
+          yield page.items.map { |row| result_item_to_hash(row) },
+            { last_evaluated_key: page.last_evaluated_key }
         end
       end
 
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
@@ -532,12 +532,15 @@ module Dynamoid
       #
       # @todo Provide support for various other options http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#query-instance_method
       def query(table_name, options = {})
-        return enum_for(:query, table_name, options) unless block_given?
-        table = describe_table(table_name)
+        Enumerator.new do |yielder|
+          table = describe_table(table_name)
 
-        Query.new(client, table, options).call.each do |page|
-          yield page.items.map { |row| result_item_to_hash(row) },
-            { last_evaluated_key: page.last_evaluated_key }
+          Query.new(client, table, options).call.each do |page|
+            yielder.yield(
+              page.items.map { |row| result_item_to_hash(row) },
+              last_evaluated_key: page.last_evaluated_key
+            )
+          end
         end
       end
 
@@ -562,13 +565,15 @@ module Dynamoid
       #
       # @todo: Provide support for various options http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#scan-instance_method
       def scan(table_name, conditions = {}, options = {})
-        return enum_for(:scan, table_name, conditions, options) unless block_given?
+        Enumerator.new do |yielder|
+          table = describe_table(table_name)
 
-        table = describe_table(table_name)
-
-        Scan.new(client, table, conditions, options).call.each do |page|
-          yield page.items.map { |row| result_item_to_hash(row) },
-            { last_evaluated_key: page.last_evaluated_key }
+          Scan.new(client, table, conditions, options).call.each do |page|
+            yielder.yield(
+              page.items.map { |row| result_item_to_hash(row) },
+              last_evaluated_key: page.last_evaluated_key
+            )
+          end
         end
       end
 
--- a/lib/dynamoid/criteria.rb
+++ b/lib/dynamoid/criteria.rb
@@ -13,12 +13,12 @@ module Dynamoid
         # see Dynamoid::Criteria::Chain.
         #
         # @since 0.2.0
-        define_method(meth) do |*args|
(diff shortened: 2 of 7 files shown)


lib/dynamoid/criteria.rb
```diff
@@ -13,12 +13,12 @@ module ClassMethods
...
         # @since 0.2.0
-        define_method(meth) do |*args|
+        define_method(meth) do |*args, &blk|
           chain = Dynamoid::Criteria::Chain.new(self)
           if args
-            chain.send(meth, *args)
+            chain.send(meth, *args, &blk)
           else
-            chain.send(meth)
+            chain.send(meth, &blk)
```
arjes: without this, the `Model.each` et al returned an enumerable even if it was passed a block.

spec/dynamoid/criteria/chain_spec.rb
```diff
@@ -936,6 +936,36 @@ def request_params
...
+          data: 'A' * 1024 * 16
+        )
+      end
+    end
+
+    it 'yields one page at a time' do
+      expect { |b| model.where(id: '1').find_by_pages(&b) }.to yield_successive_args(
+        [all(be_kind_of(model)), {last_evaluated_key: an_instance_of(Hash)}],
+        [all(be_kind_of(model)), {last_evaluated_key: nil}],
+      )
```
arjes: Actually testing the behavior here

spec/dynamoid/criteria_spec.rb
```diff
@@ -38,14 +38,16 @@
   end
 
   it 'passes each to all members' do
-    User.each do |u|
-      expect(u.id == user1.id || u.id == user2.id).to be_truthy
-      expect(u.new_record).to be_falsey
-    end
+    expect { |b| User.each(&b)}.to yield_successive_args(
+      be_kind_of(User).and(have_attributes('new_record' => false)),
+      be_kind_of(User).and(have_attributes('new_record' => false))
```
arjes: Validating the yields

lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
```diff
@@ -562,10 +563,12 @@ def query_count(table_name, options = {})
       # @todo: Provide support for various options http://docs.aws.amazon.com/sdkforruby/api/Aws/DynamoDB/Client.html#scan-instance_method
       def scan(table_name, conditions = {}, options = {})
         return enum_for(:scan, table_name, conditions, options) unless block_given?
+
         table = describe_table(table_name)
 
         Scan.new(client, table, conditions, options).call.each do |page|
-          yield page.items.map{ |row| result_item_to_hash(row) }
+          yield page.items.map { |row| result_item_to_hash(row) },
+            { last_evaluated_key: page.last_evaluated_key }
```
andrykonchin: Am I right, now we can come back to ordinal enumerator like it was before? As far as we pass two arguments to `yield`?

lib/dynamoid/criteria.rb
andrykonchin: 👍

lib/dynamoid/criteria/chain.rb
```diff
@@ -178,8 +178,8 @@ def pages_via_query
       def pages_via_scan
         return enum_for(:pages_via_scan) unless block_given?
 
-        Dynamoid.adapter.scan(source.table_name, scan_query, scan_opts).each do |items|
-          yield items.map { |hash| source.from_database(hash) }
+        Dynamoid.adapter.scan(source.table_name, scan_query, scan_opts).each do |items, metadata|
+          yield items.map { |hash| source.from_database(hash) }, metadata
```
andrykonchin: I would emphasises here we path only one particular key in `metadata` with `metadata.slice(: last_evaluated_key)`

spec/dynamoid/adapter_plugin/aws_sdk_v3_spec.rb
```diff
@@ -834,21 +837,28 @@ def dynamo_request(table_name, scan_hash = {}, select_opts = {})
...
 
-      expect(Dynamoid.adapter.scan(test_table1, name: { eq: 'Josh' })).to match_array(include({ name: 'Josh', id: '2' }, name: 'Josh', id: '1'))
+      expect(
+        Dynamoid.adapter.scan(test_table1, name: { eq: 'Josh' }).to_a
+      ).to match([
+        [
+          match_array([{ name: 'Josh', id: '2' }, { name: 'Josh', id: '1' }]),
+          {last_evaluated_key: nil}
+        ]
+      ])
```
andrykonchin: Could you please add some specs for new behavior of `query`/`scan` - returning of non-nullable `last_evaluated_key`?

spec/dynamoid/criteria_spec.rb
```diff
@@ -38,14 +38,16 @@
...
+    )
   end
 
   it 'passes find_by_pages to all members' do
-    expect(User.find_by_pages).to match([contain_exactly(user1, user2)])
+    expect {|b| User.find_by_pages(&b)}.to yield_successive_args(
+      [all(be_kind_of(User)), {last_evaluated_key: nil}],
+    )
```
andrykonchin: We need to have spec to test that returned `last_evaluated_key` is a valid argument for `start` method and paging actually works. Neither of new specs check something except it's a Hash instance.

Does it make sense?

lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
bmalinconico: Good question, I'll give it a try.

lib/dynamoid/criteria/chain.rb
bmalinconico: Sure thing

spec/dynamoid/adapter_plugin/aws_sdk_v3_spec.rb
bmalinconico: I do have that expectation at a higher level for the `find_by_pages` https://github.com/Dynamoid/dynamoid/pull/342/files#diff-de03691a59eb936547beef74a4f01686R963

I should be able to move that down to this level too.

spec/dynamoid/criteria_spec.rb
bmalinconico: Yep, I'll find somewhere to put that end-to-end(ish).

spec/dynamoid/criteria_spec.rb
```diff
@@ -38,14 +38,26 @@
...
+  end
+
+  it 'returns a last_evaluated_key which may be used to restart iteration' do
+    # Creates exactly 2 full pages
+    58.times { User.create(name: SecureRandom.uuid * 1024) }
+
+    first_page, first_page_meta = User.find_by_pages.first
+    second_page, = User.start(first_page_meta[:last_evaluated_key]).find_by_pages.first
+
+    expect(first_page & second_page).to be_empty
```
arjes: @andrykonchin This test paginates in the method described by the readme. The expectation is, a bit weird... Essentially ensuring we get  a different set of records on each call.

Is this test also sufficient for the non-null last_evaluated_key or would you like me to write one down in the adapter too?

andrykonchin: It's OK. Good job. Will review the PR a little bit later

andrykonchin:
Looks good 👍 
Thank you for the contribution

PR #343, opened 2019-04-09
Make Dynamoid::Tasks::Database a Rake::TaskLib object.

postmodern:
* This allows non-Rails users to require `dynamoid/tasks/database` and
  manually initializing the rake task via `Dynamoid::Tasks::Database.new`.
* This also allows for the rake task to become configurable via it's `#initialize` method.

### DIFF
--- a/lib/dynamoid/tasks/database.rake
+++ b/lib/dynamoid/tasks/database.rake
@@ -1,41 +1,4 @@
 # frozen_string_literal: true
 
-require 'dynamoid'
 require 'dynamoid/tasks/database'
-
-namespace :dynamoid do
-  desc 'Creates DynamoDB tables, one for each of your Dynamoid models - does not modify pre-existing tables'
-  task create_tables: :environment do
-    # Load models so Dynamoid will be able to discover tables expected.
-    Dir[File.join(Dynamoid::Config.models_dir, '**/*.rb')].sort.each { |file| require file }
-    if Dynamoid.included_models.any?
-      tables = Dynamoid::Tasks::Database.create_tables
-      result = tables[:created].map { |c| "#{c} created" } + tables[:existing].map { |e| "#{e} already exists" }
-      result.sort.each { |r| puts r }
-    else
-      puts 'Dynamoid models are not loaded, or you have no Dynamoid models.'
-    end
-  end
-
-  desc 'Tests if the DynamoDB instance can be contacted using your configuration'
-  task ping: :environment do
-    success = false
-    failure_reason = nil
-
-    begin
-      Dynamoid::Tasks::Database.ping
-      success = true
-    rescue StandardError => e
-      failure_reason = e.message
-    end
-
-    msg = "Connection to DynamoDB #{success ? 'OK' : 'FAILED'}"
-    msg << if Dynamoid.config.endpoint
-             " at local endpoint '#{Dynamoid.config.endpoint}'"
-           else
-             ' at remote AWS endpoint'
-           end
-    msg << ", reason being '#{failure_reason}'" unless success
-    puts msg
-  end
-end
+Dynamoid::Tasks::Database.new
--- a/lib/dynamoid/tasks/database.rb
+++ b/lib/dynamoid/tasks/database.rb
@@ -1,9 +1,56 @@
 # frozen_string_literal: true
 
+require 'dynamoid'
+require 'rake/tasklib'
+
 module Dynamoid
   module Tasks
-    module Database
-      module_function
+    class Database < Rake::TaskLib
+
+      # Initializes the database rake tasks.
+      def initialize
+        define
+      end
+
+      # Defines the database rake tasks.
+      def define
+        namespace :dynamoid do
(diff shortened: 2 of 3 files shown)


postmodern:
Code Climate doesn't like the overly large `define` method which contains the printing logic.

postmodern:
@andrykonchin could you give your opinion on this PR? I would really like to load the dynamodb rake tasks in non-rails projects.

andrykonchin:
Yeah, it's a good idea. Even if `dynamoid` depends on `activesupport` it could be used outside Rails application.

But it's a breaking change as far as it affects existing Rails applications. So at least it will be postponed until we have more breaking changes to release them at once (and Rails-specific staff could be moved to some small new gem like `dynamoid-rails`). 

As far as I understand the only dependency on `Rails` is depending on `:environment` task so the issue could be solved with registering empty `environment` task in not-Rails application's `Rakefile`:

```ruby
Rake::Task.define_task(:environment)
```

Does it work?

postmodern:
@andrykonchin it shouldn't break existing rails applications because the `lib/dynamoid/tasks/database.rake` file still initializes the rake tasks, albeit via `Dynamoid::Tasks::Database.new`; unless there's 3rd party code that relies on the code previously in `lib/dynamoid/tasks/database.rb`? Is `lib/dynamoid/tasks/database.rb` part of the public API or private/semi-public API?

I'm currently using padrino-rb, which also defines a common `:environment` task to load the app. [Hanami](https://guides.hanamirb.org/projects/rake/) also supports an `:environment` task. So that appears to be a defacto entry point.

andrykonchin:
Oh, I overlooked and thought the issue is in dependency on `environment` rake task.

So could you please clarify what is the issue and where is Rails-specific code that prevents using Rake tasks in non-Rails application?

postmodern:
@andrykonchin I need to load the rake tasks into a `Rakefile` of a non-Rails app. I cannot simply `require 'dynamodb/tasks/database.rake'` because `require` only loads files ending in `.rb`. Therefor, I moved the rake tasks into `dynamoid/tasks/database.rb`, and made `Dynamoid::Tasks::Database` it a sub-class of `Rake::TaskLib`. Now non-rails users can simply add the following two lines to their `Rakefile`:

    require 'dynamoid/tasks/database'
    Dynamoid::Tasks::Database.new

Note: Rails users are not impacted by this change. `dynamoid/railtie.rb` and `dynamoid/tasks/database.rake` will still define the rake tasks for them.

andrykonchin:
I see. So there is a real problem.

I see two the easiest solutions:

- just using `load` as far as it work with file extensions. I mean to write in your `Rakefile` `load 'dynamoid/tasks/database.rake'`. But it's ugly and `require` looks much better
- renaming 'dynamodb/tasks/database.rake' to 'dynamodb/tasks/database.rb' so `require` will work with it. Looks like `.rake` file extension is just a Rails convention and it shouldn't break anything (see example [here](https://github.com/jondot/sneakers/blob/master/lib/sneakers/tasks.rb)). But we already have `database.rb` file and wouldn't like to rename it or move somewhere.

Both options have cons so personally I prefer the third one. It's a common way to provide Rake tasks for non-Rails applications (at least I found several examples quickly - [sequel-rake](https://github.com/sandelius/sequel-rake/blob/master/lib/sequel/rake.rb), [dice_bag](https://github.com/mdsol/dice_bag/blob/develop/lib/dice_bag/tasks.rb) )

I propose to add file `dynamoid/tasks.rb` that will be required in any non-Rails application and to load all the `dynamoid`'s tasks there with `load`:

```ruby
# your Rakefile

require 'dynamoid/tasks'
```

```ruby
# dynamoid/tasks.rb

load 'dynamoid/tasks/database.rake'
```

What do you think about this?

postmodern:
@andrykonchin one major problem with that. `load` doesn't work the same way `require` does, mainly it expects to be given an absolute path or a path relative to the current working directory; where as `require` searches `$LOAD_PATH` for the file. Since dynamoid is installed as a gem, getting the absolute path to `dynamoid/tasks/database.rake` would involve querying rubygem's internal API. Thus, it's a common practice for rake tasks in gems to be defined in a `.rb` file as a `Rake::TaskLib` class.

postmodern:
Oh wait, it appears [load](https://ruby-doc.org/core-2.6.2/Kernel.html#method-i-load) will search `$LOAD_PATH` for the file, so that is a workaround. `dynamoid/tasks.rb` also seems reasonable.

I would still consider defining the tasks in a `Rake::TaskLib` class. This makes the tasks easier to write (no additional code in an extra `.rb` file) and easier to test.

Do other users directly use the `Dynamoid::Tasks::Database` or the `database.rb` file? If it is more part of dynamoid's private API, then I don't see any risk in changing or renaming them.

andrykonchin:
> Do other users directly use the Dynamoid::Tasks::Database or the database.rb file?

It's definitely a private API so it's not a problem to change it.

Will consider both options. Next release will include one of them.

PR #344, opened 2019-04-09
Technically {**/}*.rb is more correct/flexible.

postmodern:
Require models within the models directory, and any sub-directories.

### DIFF
--- a/lib/dynamoid/tasks/database.rake
+++ b/lib/dynamoid/tasks/database.rake
@@ -7,7 +7,7 @@ namespace :dynamoid do
   desc 'Creates DynamoDB tables, one for each of your Dynamoid models - does not modify pre-existing tables'
   task create_tables: :environment do
     # Load models so Dynamoid will be able to discover tables expected.
-    Dir[File.join(Dynamoid::Config.models_dir, '**/*.rb')].sort.each { |file| require file }
+    Dir[File.join(Dynamoid::Config.models_dir, '{**/}*.rb')].sort.each { |file| require file }
     if Dynamoid.included_models.any?
       tables = Dynamoid::Tasks::Database.create_tables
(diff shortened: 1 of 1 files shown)


andrykonchin:
Could you please point at documentation where this difference is explained or illustrated? I can't find any mention about `{}` in this context in path pattern ([Dir.glob](https://ruby-doc.org/core-2.6/Dir.html#glob-method), [File.fnmatch](https://ruby-doc.org/core-2.6/File.html#method-c-fnmatch))

postmodern:
https://www.thoughtco.com/using-glob-with-directories-2907832

> * – Match zero or more characters. A glob consisting of only the asterisk and no other characters or wildcards will match all files in the current directory. The asterisk is usually combined with a file extension if not more characters to narrow down the search.
> ** – Match all directories recursively. This is used to descend into the directory tree and find all files in sub-directories of the current directory, rather than just files in the current directory. This wildcard is explored in the example code below.
> ? – Match any one character. This is useful for finding files whose name are in a particular format. For example, 5 characters and a .xml extension could be expressed as ?????.xml.
> [a-z] – Match any character in the character set. The set can be either a list of characters or a range separated with the hyphen character. Character sets follow the same syntax as and behave in the same manner as character sets in regular expressions.
> {a,b} – Match pattern a or b. Though this looks like a regular expression quantifier, it isn't. For example, in regular expression, the pattern a{1,2} will match 1 or 2 'a' characters. In globbing, it will match the string a1 or a2. Other patterns can be nested inside of this construct.

So if you wanted to search in `.` and all sub-directories, you'd use `{.,**/}*.rb`, which can then be further simplified to `{**/}*.rb`.

postmodern:
Er wait, I made a mistake. It should be `{,**/}*.rb`...

PR #346, opened 2019-04-09
Use the `{,**/}*.rb` dir glob which is technically more correct.

postmodern:
Technically, `**/*.rb` will match any `.rb` file in the directory, as well as any sub-directories. However, `{,**/}*.rb` is more explicit, as the `{,b}` pattern expands to "nothing OR `b`".


postmodern:
Hmm, second thoughts, this isn't really an improvement.

PR #347, opened 2019-04-12
Use the official amazon/dynamodb-local image

postmodern:
Switch to using the official Amazon dynamodb-local image.

https://hub.docker.com/r/amazon/dynamodb-local/


andrykonchin:
Thank you for the contribution
👍

PR #362, opened 2019-06-10
Handle edge case where there is a GSI with the same hash key as primary

atyndall:
In the current implementation, when a `where` query is performed with a hash key and range key that would match a GSI, but the hash key overlaps with the primary index, the GSI is not selected and an expensive Scan is performed instead of the cheap Query.

I ran into this because I wanted to create an index on a different range key without reindexing all the data. This meant the easiest solution was for me to create a GSI with the same hash key as the primary index, but with a different range key. I then encountered this bug.


andrykonchin:
Looks good 👍 

Thank you for the contribution.

PR #370, opened 2019-07-27
add precision option for datetime string dumping

f0ster:


andrykonchin:
Great, but the feature is missing specs. Please add specs for `Dynamoid::Dumping` module.

f0ster:
@andrykonchin @pboling added specs

lib/dynamoid/type_casting.rb
```diff
@@ -227,14 +227,22 @@ def process(value)
           nil
         elsif value.is_a?(String)
           dt = begin
+                 # parse first for stricter constraint failure before attempting to parse iso for precision support
+                 # e.g. DateTime.iso8601("2018-12") is valid
                  DateTime.parse(value)
+                 DateTime.iso8601(value)
```
andrykonchin: I don't think we need these changes in type casting. Anything that is like date/time could be casted to DateTime. Type casting isn't connected with writing/reading attributes to/from DynamoDB.

lib/dynamoid/undumping.rb
```diff
@@ -202,7 +202,7 @@ def process(value)
                             else
                               @options[:store_as_string]
                             end
-        value = DateTime.iso8601(value).to_time.to_i if use_string_format
+        value = DateTime.iso8601(value).to_time if use_string_format
```
andrykonchin: I believe it should be tested as well. It's a new feature - storing milliseconds when use string format for datetime.

spec/dynamoid/dumping_spec.rb
```diff
@@ -239,6 +239,8 @@
         Dynamoid.config.store_datetime_as_string = store_datetime_as_string
       end
 
+
+
```
andrykonchin: Excessive new lines detected

lib/dynamoid/type_casting.rb
f0ster: @andrykonchin Per the comment,  the existing expected behaviour is to break for poorly parsed strings.   https://github.com/Dynamoid/dynamoid/blob/master/spec/dynamoid/type_casting_spec.rb#L143

I wrote this work around to support precision, and also keep the same erroneous expectations.  Parse is expected to throw an exception, where as constructing with iso8601 has less strict input expectations.

spec/dynamoid/dumping_spec.rb
```diff
@@ -366,6 +368,18 @@
...
+      it 'stores in string format with precision when :iso_precision is present' do
+        klass = new_class do
+          field :signed_up_on, :datetime, store_as_string: true, iso_precision: 6
+        end
+
+        signed_up_on = DateTime.now.utc.iso8601(6)
+
+        obj = klass.create(signed_up_on: signed_up_on)
+
+        expect(reload(obj).signed_up_on.to_f).to eql(DateTime.iso8601(signed_up_on).to_f)
```
andrykonchin: I think i't completely OK to use `to_f` here but it's a bad practice to check equality of floating point numbers. Compare milliseconds or string representation would be more correct.

The precision is 6 in the test so it's OK, but with precision 9 (with nanoseconds) this test could become flaky because MRI isn't accurate in preserving nanoseconds for Time/DateTime.

lib/dynamoid/undumping.rb
f0ster: I agree, I think this could be a clearer abstraction. FWIW, This was amended to make the current and additional spec coverage pass

lib/dynamoid/type_casting.rb
andrykonchin: I see. Well. 
```ruby
                 DateTime.parse(value)
                 DateTime.iso8601(value)
```
Here. `DateTime.parse` supports precision, doesn't it?
```ruby
irb(main):003:0> s = DateTime.now.iso8601(2)
=> "2019-07-29T22:06:24.62+03:00"
irb(main):004:0> DateTime.parse(s).usec
=> 620000
```
Here 
```ruby
            if dt_has_precision
              dt.utc
```

completely ignores `ApplicationTimeZone.utc_offset` what's wrong. If passed string without timezone specified explicitly `ApplicationTimeZone.utc_offset` should be used.

I thinks precision support could be added directly in this line:
```ruby
DateTime.new(dt.year, dt.mon, dt.mday, dt.hour, dt.min, dt.sec, offset)
```
seconds (dt.sec) could have fraction so this line could be changed in this way:
```ruby
DateTime.new(dt.year, dt.mon, dt.mday, dt.hour, dt.min, dt.se + dt.sec_fraction, offset)
```

Look at the example:
```ruby
irb(main):015:0> s = DateTime.now.iso8601(4)
=> "2019-07-29T22:24:20.8163+03:00"
irb(main):016:0> DateTime.parse(s).sec
=> 20
irb(main):018:0> DateTime.parse(s).sec_fraction
=> (8163/10000)
```

andrykonchin:
@f0ster Did you have a chance to look through my comments?

PR #382, opened 2019-09-20
Fix deprecation warning about Module#parent_name in Rails 6

tmandke:


andrykonchin:
Great.

Have you considered using `String#deconstantize method`? Wouldn't the code be simpler without checking Rails version?

Could you fix failing specs?

tmandke:
> Have you considered using String#deconstantize method?

I don't understand how the applies here. I was trying to keep it as is while dropping the deprecation warning.

> Could you fix failing specs?

 Will do

andrykonchin:
Good job 👍

PR #388, opened 2019-10-27
Allocate only one hash to initialize the attributes of the document

arjes:
Currently the initialize is allocating a LOT of objects which are unnecessary. 

The first object is allocated for reduce and then on each attribute which is passed in, or has a default, a hash is allocated for the argument to merge, and then another hash returned from the merge call for the next iteration.

This change allocates only a single hash for the entire initialization process. 

No spec changes since this is expected to have no functional change


andrykonchin:
Thank you for this improvement 👍

PR #393, opened 2019-11-05
ISSUE-391: support pre-configured credentials passed in

emmajhyde:
Provides support for pre-configured credentials being passed into the client, as in: 
```ruby
@client = Aws::DynamoDB::Client.new(region: _region, credentials: _credentials)
```

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
@@ -75,11 +75,18 @@ module Dynamoid
         (Dynamoid::Config.settings.compact.keys & CONNECTION_CONFIG_OPTIONS).each do |option|
           @connection_hash[option] = Dynamoid::Config.send(option)
         end
-        if Dynamoid::Config.access_key?
-          @connection_hash[:access_key_id] = Dynamoid::Config.access_key
-        end
-        if Dynamoid::Config.secret_key?
-          @connection_hash[:secret_access_key] = Dynamoid::Config.secret_key
+
+        # if credentials are passed, they already contain access key & secret key
+        if Dynamoid::Config.credentials?
+          connection_hash[:credentials] = Dynamoid::Config.credentials
+        else
+          # otherwise, pass access key & secret key for credentials creation
+          if Dynamoid::Config.access_key?
+            connection_hash[:access_key_id] = Dynamoid::Config.access_key
+          end
+          if Dynamoid::Config.secret_key?
+            connection_hash[:secret_access_key] = Dynamoid::Config.secret_key
+          end
         end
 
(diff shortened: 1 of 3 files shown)


emmajhyde:
Wasn't sure where to test for this. Any direction is appreciated!

lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
```diff
@@ -92,6 +88,21 @@ def connection_config
...
+        else
+          # otherwise, pass access key & secret key for credentials creation
+          if Dynamoid::Config.access_key?
+            connection_hash[:access_key_id] = Dynamoid::Config.access_key
+          end
+          if Dynamoid::Config.secret_key?
+            connection_hash[:secret_access_key] = Dynamoid::Config.secret_key
+          end
+        end
+      end
```
emmajhyde: Made a new method - tooling complained about complexity. Open to other ideas!

emmajhyde: Ah, moving it out didn't reduce the method complexity. I guess this alone has a complexity of 6. Let me know how I can remedy this.

emmajhyde:
@andrykonchin this is regarding the issue I brought up with you last week. Could I get a review so I know where to test and how to remedy the codeclimate failure?

andrykonchin:
First of all - good job 👍 and sorry for the delay.

Issue with Code Climate - I think the warning doesn't matter and I propose to keep the first simplest version without adding new method (I suppose it would be easier to refactor later).

Testing - as you can see configuration is not tested well (there are only tests for added recently backoff strategies). So it isn't an issue to skip adding tests now. I plan to cover fully configuration with tests soon.

The only thing I would add is documentation. There is a list of config options in the Readme file - we need to update it. And it would be great to have an example of configuring credentials with either `Credentials`/`SharedCredentials`/... in Readme and/or in a code as a comment.

emmajhyde:
Let me know how this look @andrykonchin !

andrykonchin:
Thank you for the contribution!

zerooverride:
It seems this change has broken the gem, when I update to 3.4.0 I get this error (regardless of using the new feature or not):

NameError: undefined local variable or method `connection_hash' for #<Dynamoid::AdapterPlugin::AwsSdkV3:0x0782cda8>
Did you mean?  @connection_hash

That looks like a valid error and changing aws_sdk_v3.rb to always use @connection_hash makes the change seem to work as expected.

Am I maybe doing something wrong or does this need to be fixed?

emmajhyde:
@zerooverride, you're absolutely right, I had originally made a separate connection_hash method to try to bypass an ABC Complexity Code Climate warning, and forgot to switch it back. My bad! Thank you for posting. I've made a PR here: 

https://github.com/Dynamoid/dynamoid/pull/398

andrykonchin:
@zerooverride 

Thanks for reporting the issue. It's fixed on the master. Will release it soon.

PR #398, opened 2019-12-01
ISSUE-391: broke build - fix for credentials impl

emmajhyde:
Regarding: https://github.com/Dynamoid/dynamoid/pull/393
Addresses failure reported in above link by @zerooverride: 
"It seems this change has broken the gem, when I update to 3.4.0 I get this error (regardless of using the new feature or not):

NameError: undefined local variable or method `connection_hash' for #Dynamoid::AdapterPlugin::AwsSdkV3:0x0782cda8
Did you mean? @connection_hash

That looks like a valid error and changing aws_sdk_v3.rb to always use @connection_hash makes the change seem to work as expected.

Am I maybe doing something wrong or does this need to be fixed?"


andrykonchin:
Great. Thanks

PR #408, opened 2020-01-28
add active support load hooks

aaronmallen:
Add `ActiveSupport` load hook on `Dynamoid::Document` load

### DIFF
--- a/lib/dynamoid/document.rb
+++ b/lib/dynamoid/document.rb
@@ -225,3 +225,5 @@ module Dynamoid #:nodoc:
     end
   end
 end
+
+ActiveSupport.run_load_hooks(:dynamoid, Dynamoid::Document)
--- a/spec/dynamoid/document_spec.rb
+++ b/spec/dynamoid/document_spec.rb
@@ -3,6 +3,12 @@
 require 'spec_helper'
 
 describe Dynamoid::Document do
+  it 'runs load hooks' do
+    result = nil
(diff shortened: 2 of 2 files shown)


andrykonchin:
Looks good. Can you add some specs?

aaronmallen:
> Looks good. Can you add some specs?

Sorry, this is serious or a joke request?

aaronmallen:
I'm not sure how to go about writing a spec for this... I looked through rails/rails and heartcombo/devise to see how they may have tested it and they don't.  The problem being when `spec_helper` is loaded `Dynamoid::Document` is already loaded so there isn't away I'm aware of to spy on `ActiveSupport` before this happens.  I could try to unload `Dynomoid::Document` and re-require it but I'm not sure what kind of side effects that would introduce to the spec.  Do you have any suggestions on what kind of spec you'd like to see here?

andrykonchin:
If I understand correctly the source code here

https://github.com/rails/rails/blob/6-0-stable/activesupport/lib/active_support/lazy_load_hooks.rb#L41-L54

registering a new hook (with `on_load`) even when `run_load_hooks` method call already happened still runs new hook. So it could be tested this way - register a hook and checking it was immediately called.

I am afraid if there is no any proper test this functionality could be easily broken/removed in Dynamoid without understanding what it is.

andrykonchin:
Looks good 👍 
Thanks

PR #410, opened 2020-02-06
Don't specify provisioned capacity in table creation for on-demand

### LINKED ISSUE #409: GSI with on-demand billing fails to create table
Trying to create a table fails when the table uses on-demand capacity and a global secondary index. AWS reports the following:
```
One or more parameter values were invalid: ProvisionedThroughput should not be specified for index: app-dev_content_index_type_id when BillingMode is PAY_PER_REQUEST
```

A simple class that reproduces the issue:
```ruby
class ApplicationItem
  include Dynamoid::Document

  table name: :content,
        key: :id,
        capacity_mode: :on_demand,
        timestamps: false

  field :type # For STI.

  global_secondary_index hash_key: :type, range_key: :id, projected_attributes: :all
end
```

The request to create the table that was generated by Dynamoid:
```json
{
  "TableName": "app-dev_content",
  "KeySchema": [
    {
      "AttributeName": "id",
      "KeyType": "HASH"
    }
  ],
  "AttributeDefinitions": [
    {
      "AttributeName": "id",
      "AttributeType": "S"
    },
    {
      "AttributeName": "type",
      "AttributeType": "S"
    }
  ],
  "BillingMode": "PAY_PER_REQUEST",
  "GlobalSecondaryIndexes": [
    {
      "IndexName": "app-dev_content_index_type_id",
      "KeySchema": [
        {
          "AttributeName": "type",
          "KeyType": "HASH"
        },
        {
          "AttributeName": "id",
          "KeyType": "RANGE"
        }
      ],
      "Projection": {
        "ProjectionType": "ALL"
      },
      "ProvisionedThroughput": {
        "ReadCapacityUnits": 100,
        "WriteCapacityUnits": 20
      }
    }
  ]
}
```

icy-arctic-fox:
Should fix #409

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3/create_table.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3/create_table.rb
@@ -213,7 +213,7 @@ module Dynamoid
           end
 
           # Only global secondary indexes have a separate throughput.
-          if index.type == :global_secondary
+          if index.type == :global_secondary && options[:billing_mode] != :on_demand
             hash[:provisioned_throughput] = {
               read_capacity_units: index.read_capacity,
               write_capacity_units: index.write_capacity
--- a/spec/dynamoid/adapter_plugin/aws_sdk_v3/create_table_spec.rb
+++ b/spec/dynamoid/adapter_plugin/aws_sdk_v3/create_table_spec.rb
@@ -1,0 +1,192 @@
+# frozen_string_literal: true
+
+require 'dynamoid/adapter_plugin/aws_sdk_v3/create_table'
+require 'spec_helper'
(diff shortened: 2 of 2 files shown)


andrykonchin:
Looks good. Could you add specs for the fix?

Dynamodb-local doesn't behave like real DynamoDB and doesn't return the error, so we can add only a unit test. There is no proper place for isolated specs for create table operation - it's tested with dynamodb-local. So I believe we need new spec for the `CreateTable` class.

icy-arctic-fox:
Ok, I added some specs for the `CreateTable` class. The last one in the file tests the issue.

andrykonchin:
Great. Thank you.

Issue #217, opened 2018-01-09
Question: How to ensure clean state for testing


pratik60:
Hey!

We started prototyping using Dynamoid, and are really liking the experience compared to aws-record.

One question I have is how can we clean the table after running a test. I have some tests which do check the count of items in a table, and currently have to resort to a delete_table, and create_table in before(:each) block. 

Is there a better way to do this? Can I issue a delete_all in a table?

For reference most companies use https://github.com/DatabaseCleaner/database_cleaner for mysql, postgres, mongoid etc

pboling:
There should be a wiki about testing with Dynamoid, but for now thanks for asking!

Near the top of my `spec/rails_helper.rb`:
```
require 'support/dynamoid'
```

Then later:
```
  config.before(:example, :active_record_clean) do
    DatabaseCleaner[:active_record].strategy = :transaction
  end
  config.before(:example, :mongoid_clean) do
    DatabaseCleaner[:mongoid].strategy = :truncation
  end
  config.before(:example, :redis_clean) do
    DatabaseCleaner[:redis].strategy = :truncation
  end
  config.before(:example, :dynamoid_clean) do
    DynamoidReset.all
  end
```
Note the second argument to the calls to `:before` above.  Those are spec tags, and you use them to mark specific tests as needing a particular type of cleaning.  The nice thing about this setup is your specs end up defining the edges of you data layer dependencies.

I have in `spec/support/dynamoid.rb`:
```
module DynamoidReset
  def self.all
    raise RuntimeError unless Rails.env.test?
    Dynamoid.adapter.list_tables.each do |table|
      # Only delete tables in our namespace
      if table =~ /^#{Dynamoid::Config.namespace}/
        Dynamoid.adapter.delete_table(table)
      end
    end
    Dynamoid.adapter.tables.clear
    # Recreate all tables to avoid unexpected errors
    Dynamoid.included_models.each(&:create_table)
  end
end

# Reduce noise in test output
Dynamoid.logger.level = Logger::FATAL
```

Then a spec might look like:
```
RSpec.describe `A thing` do
  it 'works', :active_record_clean, :mongoid_clean, :redis_clean, :dynamoid_clean do
    expect('all the things').to eq('all the things')
  end
end
```

Those spec tags can go at any level, top describe, middle describe or context, or individual spec definitions.

andrykonchin:
Look at a Test Environment section in the Readme (https://github.com/Dynamoid/Dynamoid#test-environment)

[...]

PR #416, opened 2020-02-28
Use #batch_delete_item in #truncate

TheSmartnik:
Hi! I thought I'd rewrite truncate method as was suggested in this comment https://github.com/Dynamoid/dynamoid/issues/217#issuecomment-568222683


andrykonchin:
Looks good 👍 Thank you

PR #425, opened 2020-04-17
fix: removes DynamoId to Dynamoid typos in README.md

omarsotillo:
Starting with this project by checking the documentation in the README.md, there are some typos that could lead to debugging why the gem is not being loaded.


andrykonchin:
Looks great. Thank you.

PR #432, opened 2020-04-28
Support tables that use "hash_key" as their partition key name

remomueller:
Working with tables that have their partition key set to "hash_key" causes a "stack level too deep" loop when looking up the symbol to reference the value. This CR updates the instances to make use `self` instead of `send` to look up the instance property when available, instead of making a call to the same instance method.


andrykonchin:
Good job 👍 . Thank you for the fix.

remomueller:
Thank you!

PR #440, opened 2020-05-27
Add record_limit to Criteria::Chain#first

mrkamel:
Hi, thanks for this nice gem. I recognized `Criteria::Chain#first` basically fetches all records matching the criteria and then returns just the first document. This is rather inefficient, such that this PR applies `record_limit(1)` within `#first`. However, i'm still a bit new to dynamo/dynamoid, so maybe i'm not yet aware of any complications this could introduce.

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -176,6 +176,25 @@ module Dynamoid
         end
       end
 
+      # Returns the first item matching the criteria.
+      #
+      #   Post.where(links_count: 2).first
+      #
+      # Applies `record_limit(1)` to ensure only a single record is fetched
+      # when no non-key conditions are present.
+      #
+      # If used without criteria it just returns the first item of some
+      # arbitrary order.
+      #
+      #   Post.first
+      #
+      # @return [Model|nil]
+      def first
+        return super if @key_fields_detector.non_key_present?
+
+        record_limit(1).to_a.first
+      end
+
       # Returns the last item matching the criteria.
       #
       #   Post.where(links_count: 2).last
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -181,7 +181,8 @@ module Dynamoid
       #   Post.where(links_count: 2).first
       #
       # Applies `record_limit(1)` to ensure only a single record is fetched
-      # when no non-key conditions are present.
+      # when no non-key conditions are present and `scan_limit(1)` when no
+      # conditions are present at all.
       #
       # If used without criteria it just returns the first item of some
       # arbitrary order.
@@ -190,6 +191,7 @@ module Dynamoid
       #
       # @return [Model|nil]
       def first
+        return scan_limit(1).to_a.first if @query.blank?
         return super if @key_fields_detector.non_key_present?
 
         record_limit(1).to_a.first
--- a/lib/dynamoid/criteria/key_fields_detector.rb
+++ b/lib/dynamoid/criteria/key_fields_detector.rb
@@ -10,6 +10,10 @@ module Dynamoid #:nodoc:
           @fields = query_hash.keys.map(&:to_s).map { |s| s.split('.').first }
         end
 
+        def contain_only?(field_names)
+          (@fields - field_names.map(&:to_s)).blank?
+        end
+
         def contain?(field_name)
           @fields.include?(field_name.to_s)
         end
@@ -24,6 +28,10 @@ module Dynamoid #:nodoc:
         @source = source
         @query = Query.new(query)
         @result = find_keys_in_query
+      end
+
+      def non_key_present?
+        !@query.contain_only?([hash_key, range_key].compact)
       end
 
       def key_present?
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -1430,6 +1430,80 @@ describe Dynamoid::Criteria::Chain do
     end
   end
 
+  describe '#first' do
+    let(:model) do
+      new_class partition_key: :name do
+        range :age, :integer
+        field :city, :string
+      end
+    end
+
+    it 'applies a record_limit' do
+      model.create(name: 'Bob', age: 5)
+
+      chain = Dynamoid::Criteria::Chain.new(model)
+      expect(chain).to receive(:record_limit).with(1)
+      chain.first
+    end
+
+    it 'applies a record limit if only key conditions are present' do
+      model.create(name: 'Bob', age: 5)
+
+      chain = Dynamoid::Criteria::Chain.new(model)
+      expect(chain).to receive(:record_limit).with(1)
+      chain.where(name: 'Bob', age: 5).first
+    end
+
+    it 'does not apply a record limit if the hash key is missing' do
+      model.create(name: 'Bob', city: 'New York', age: 5)
+
+      chain = Dynamoid::Criteria::Chain.new(model)
+      expect(chain).not_to receive(:record_limit)
+      chain.where(age: 5).first
+    end
+
+    it 'does not apply a record limit if non-key conditions are present' do
+      model.create(name: 'Bob', city: 'New York', age: 5)
+
+      chain = Dynamoid::Criteria::Chain.new(model)
+      expect(chain).not_to receive(:record_limit)
+      chain.where(city: 'New York').first
+      chain.where(name: 'Bob', city: 'New York').first
+      chain.where(name: 'Bob', age: 5, city: 'New York').first
+    end
+
+    it 'does not apply a record limit if non-equality conditions are present' do
+      model.create(name: 'Bob', age: 5)
+      model.create(name: 'Alice', age: 6)
+
+      chain = Dynamoid::Criteria::Chain.new(model)
+      expect(chain).not_to receive(:record_limit)
+      chain.where('name.gt': 'Alice').first
+    end
+
+    it 'returns nil if no matching document is present' do
+      model.create(name: 'Bob', age: 5)
+
+      expect(model.where(name: 'Alice').first).to be_nil
+    end
+
+    it 'returns the first document with regards to the sort order' do
+      customer1 = model.create(name: 'Bob', age: 5)
+      customer2 = model.create(name: 'Bob', age: 9)
+      customer3 = model.create(name: 'Bob', age: 12)
+
+      expect(model.first.age).to eq(5)
+    end
+
+    it 'returns the first document matching the criteria and with regards to the sort order' do
+      customer1 = model.create(name: 'Bob', age: 4)
+      customer3 = model.create(name: 'Alice', age: 6)
+      customer4 = model.create(name: 'Alice', age: 8)
+
+      expect(model.where(name: 'Alice').first.age).to eq(6)
+    end
+  end
+
   describe '#count' do
     describe 'Query vs Scan' do
       it 'Scans when query is empty' do
--- a/spec/dynamoid/criteria/chain_spec.rb
+++ b/spec/dynamoid/criteria/chain_spec.rb
@@ -1438,47 +1438,47 @@ describe Dynamoid::Criteria::Chain do
       end
     end
 
-    it 'applies a record_limit' do
-      model.create(name: 'Bob', age: 5)
+    it 'applies a scan limit if no conditions are present' do
+      document = model.create(name: 'Bob', age: 5)
 
       chain = Dynamoid::Criteria::Chain.new(model)
-      expect(chain).to receive(:record_limit).with(1)
-      chain.first
+      expect(chain).to receive(:scan_limit).with(1).and_call_original
+      expect(chain.first).to eq(document)
     end
 
     it 'applies a record limit if only key conditions are present' do
-      model.create(name: 'Bob', age: 5)
+      document = model.create(name: 'Bob', age: 5)
 
       chain = Dynamoid::Criteria::Chain.new(model)
-      expect(chain).to receive(:record_limit).with(1)
-      chain.where(name: 'Bob', age: 5).first
+      expect(chain).to receive(:record_limit).with(1).and_call_original
+      expect(chain.where(name: 'Bob', age: 5).first).to eq(document)
     end
 
     it 'does not apply a record limit if the hash key is missing' do
-      model.create(name: 'Bob', city: 'New York', age: 5)
+      document = model.create(name: 'Bob', city: 'New York', age: 5)
 
       chain = Dynamoid::Criteria::Chain.new(model)
       expect(chain).not_to receive(:record_limit)
-      chain.where(age: 5).first
+      expect(chain.where(age: 5).first).to eq(document)
     end
 
     it 'does not apply a record limit if non-key conditions are present' do
-      model.create(name: 'Bob', city: 'New York', age: 5)
+      document = model.create(name: 'Bob', city: 'New York', age: 5)
 
       chain = Dynamoid::Criteria::Chain.new(model)
       expect(chain).not_to receive(:record_limit)
-      chain.where(city: 'New York').first
-      chain.where(name: 'Bob', city: 'New York').first
-      chain.where(name: 'Bob', age: 5, city: 'New York').first
+      expect(chain.where(city: 'New York').first).to eq(document)
+      expect(chain.where(name: 'Bob', city: 'New York').first).to eq(document)
+      expect(chain.where(name: 'Bob', age: 5, city: 'New York').first).to eq(document)
     end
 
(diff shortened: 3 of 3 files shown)


andrykonchin:
Actually it's a very good idea but there is only one issue.

The provided implementation works only when there are only conditions on primary key fields (hash key `name` and range key `age` in the specs) and doesn't work properly for more general case:

```ruby
new_class do
  field :name
  field :age, :integer
  field :dob, :date
end
```

Let's run the one of the provided specs:

```ruby
it 'returns the first document matching the criteria and with regards to the sort order' do
  customer1 = model.create(name: 'Bob', age: 4)
  customer3 = model.create(name: 'Alice', age: 6)
  customer4 = model.create(name: 'Alice', age: 8)

  expect(model.where(name: 'Alice').first.age).to eq(6)
end
```

It fails (running against dynamodb-local) because there are two items with name = 'Alice' and only one of them will be read. Actually any of them randomly I believe.

```
  1) Dynamoid::Criteria::Chain#first returns the first document matching the criteria and with regards to the sort order
     Failure/Error: expect(model.where(name: 'Alice').first.age).to eq(6)

       expected: 6
            got: 8

       (compared using ==)
```

So I would say it's a completely reasonable optimization but it should be limited and be applied only when there are conditions only on hash and range keys.

Another subtle issue - GSI or LSI can be chosen and they have different hash/range keys. So it should also be taken into account.

mrkamel:
Thanks for your quick feedback!

Hm, but in your example you don't have any keys defined for the table (hash/range). So i'd argue the spec simply can't expect that age 6 will be returned, as any record could be returned - and could be returned before my change as well?!

In my specs, however, age is a range key, such that the first doc returned should be the one with lowest age.

Regarding GSI/LSI ... basically the same thing. For scan queries (e.g. GSI without projected_attributes: all), no order is guaranteed, such that no order can be expected.

andrykonchin:
Yeah, you are right. My example is completely wrong. Moreover I confused `record_limit` with `scan_limit` so all my arguments are incorrect.

But nevertheless `record_limit` has own issues. Using it with limit 1 is the worth case. Internally a [Limit](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-Limit) parameter is used.

Let's consider an example with condition on a hash key and a range key (like in the provided specs).  Dynamoid sends a Query request with `Limit=1` and with key conditions. DynamoDB reads only one item with specified hash key and range key.

Let's consider another example where there is additional condition on some non-key field.  Dynamoid sends a Query request with `Limit=1`, with key conditions and non-key conditions. DynamoDB reads only one item with specified hash key and range key and check non-key condition. If an item matches the condition - DynamoDB returns in. Otherwise it returns empty result. When Dynamoid receives empty result it send another Query request with the same conditions, `Limit` parameter but with [LastEvaluatedKey](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-response-LastEvaluatedKey) to skip already read items. Dynamoid will send Query requests to read 1 item one by one until DynamoDB reads an items which matches a non-key condition.

Let's illustrate it with the following test (on current `master` branch):

```ruby
    let(:model) do
      new_class partition_key: :name do
        range :age, :integer
        field :amount, :integer
      end
    end

    it 'returns the first document matching the criteria and with regards to the sort order', log_level: :debug do
      customer1 = model.create(name: 'Bob',   age: 4, amount: 1)
      customer3 = model.create(name: 'Alice', age: 6, amount: 2)
      customer3 = model.create(name: 'Alice', age: 7, amount: 2)
      customer4 = model.create(name: 'Alice', age: 8, amount: 3)

      expect(model.where(name: 'Alice', amount: 3).record_limit(1).first.age).to eq(8)
    end
```

It will produce following requests to DynamoDB:

<details><summary>Click me</summary>

```
D, [2020-05-28T20:57:28.677326 #35684] DEBUG -- : CreateTable

Request:

{
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "KeySchema": [
    {
      "AttributeName": "name",
      "KeyType": "HASH"
    },
    {
      "AttributeName": "age",
      "KeyType": "RANGE"
    }
  ],
  "AttributeDefinitions": [
    {
      "AttributeName": "name",
      "AttributeType": "S"
    },
    {
      "AttributeName": "age",
      "AttributeType": "N"
    }
  ],
  "BillingMode": "PROVISIONED",
  "ProvisionedThroughput": {
    "ReadCapacityUnits": 100,
    "WriteCapacityUnits": 20
  }
}

Response:

{
  "TableDescription": {
    "AttributeDefinitions": [
      {
        "AttributeName": "name",
        "AttributeType": "S"
      },
      {
        "AttributeName": "age",
        "AttributeType": "N"
      }
    ],
    "TableName": "dynamoid_tests_documents_1590688648_522",
    "KeySchema": [
      {
        "AttributeName": "name",
        "KeyType": "HASH"
      },
      {
        "AttributeName": "age",
        "KeyType": "RANGE"
      }
    ],
    "TableStatus": "ACTIVE",
    "CreationDateTime": 1590688648.699,
    "ProvisionedThroughput": {
      "LastIncreaseDateTime": 0.0,
      "LastDecreaseDateTime": 0.0,
      "NumberOfDecreasesToday": 0,
      "ReadCapacityUnits": 100,
      "WriteCapacityUnits": 20
    },
    "TableSizeBytes": 0,
    "ItemCount": 0,
    "TableArn": "arn:aws:dynamodb:ddblocal:000000000000:table/dynamoid_tests_documents_1590688648_522"
  }
}

D, [2020-05-28T20:57:28.683241 #35684] DEBUG -- : PutItem

Request:

{
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "Item": {
    "name": {
      "S": "Bob"
    },
    "age": {
      "N": "4"
    },
    "amount": {
      "N": "1"
    },
    "created_at": {
      "N": "1590688648.677717"
    },
    "updated_at": {
      "N": "1590688648.677784"
    }
  },
  "Expected": {
    "name": {
      "Exists": false
    },
    "age": {
      "Exists": false
    }
  }
}

Response:

{
}

D, [2020-05-28T20:57:28.688127 #35684] DEBUG -- : PutItem

Request:

{
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "Item": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "6"
    },
    "amount": {
      "N": "2"
    },
    "created_at": {
      "N": "1590688648.683663"
    },
    "updated_at": {
      "N": "1590688648.6837"
    }
  },
  "Expected": {
    "name": {
      "Exists": false
    },
    "age": {
      "Exists": false
    }
  }
}

Response:

{
}

D, [2020-05-28T20:57:28.692821 #35684] DEBUG -- : PutItem

Request:

{
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "Item": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "7"
    },
    "amount": {
      "N": "2"
    },
    "created_at": {
      "N": "1590688648.688552"
    },
    "updated_at": {
      "N": "1590688648.688589"
    }
  },
  "Expected": {
    "name": {
      "Exists": false
    },
    "age": {
      "Exists": false
    }
  }
}

Response:

{
}

D, [2020-05-28T20:57:28.698356 #35684] DEBUG -- : PutItem

Request:

{
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "Item": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "8"
    },
    "amount": {
      "N": "3"
    },
    "created_at": {
      "N": "1590688648.693225"
    },
    "updated_at": {
      "N": "1590688648.693261"
    }
  },
  "Expected": {
    "name": {
      "Exists": false
    },
    "age": {
      "Exists": false
    }
  }
}

Response:

{
}

D, [2020-05-28T20:57:28.735726 #35684] DEBUG -- : DescribeTable

Request:

{
  "TableName": "dynamoid_tests_documents_1590688648_522"
}

Response:

{
  "Table": {
    "AttributeDefinitions": [
      {
        "AttributeName": "name",
        "AttributeType": "S"
      },
      {
        "AttributeName": "age",
        "AttributeType": "N"
      }
    ],
    "TableName": "dynamoid_tests_documents_1590688648_522",
    "KeySchema": [
      {
        "AttributeName": "name",
        "KeyType": "HASH"
      },
      {
        "AttributeName": "age",
        "KeyType": "RANGE"
      }
    ],
    "TableStatus": "ACTIVE",
    "CreationDateTime": 1590688648.699,
    "ProvisionedThroughput": {
      "LastIncreaseDateTime": 0.0,
      "LastDecreaseDateTime": 0.0,
      "NumberOfDecreasesToday": 0,
      "ReadCapacityUnits": 100,
      "WriteCapacityUnits": 20
    },
    "TableSizeBytes": 237,
    "ItemCount": 4,
    "TableArn": "arn:aws:dynamodb:ddblocal:000000000000:table/dynamoid_tests_documents_1590688648_522"
  }
}

D, [2020-05-28T20:57:28.740349 #35684] DEBUG -- : Query

Request:

{
  "ConsistentRead": false,
  "ScanIndexForward": true,
  "Limit": 1,
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "KeyConditions": {
    "name": {
      "ComparisonOperator": "EQ",
      "AttributeValueList": [
        {
          "S": "Alice"
        }
      ]
    }
  },
  "QueryFilter": {
    "amount": {
      "ComparisonOperator": "EQ",
      "AttributeValueList": [
        {
          "N": "3"
        }
      ]
    }
  }
}

Response:

{
  "Items": [

  ],
  "Count": 0,
  "ScannedCount": 1,
  "LastEvaluatedKey": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "6"
    }
  }
}

D, [2020-05-28T20:57:28.744816 #35684] DEBUG -- : Query

Request:

{
  "ConsistentRead": false,
  "ScanIndexForward": true,
  "Limit": 1,
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "KeyConditions": {
    "name": {
      "ComparisonOperator": "EQ",
      "AttributeValueList": [
        {
          "S": "Alice"
        }
      ]
    }
  },
  "QueryFilter": {
    "amount": {
      "ComparisonOperator": "EQ",
      "AttributeValueList": [
        {
          "N": "3"
        }
      ]
    }
  },
  "ExclusiveStartKey": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "6.0"
    }
  }
}

Response:

{
  "Items": [

  ],
  "Count": 0,
  "ScannedCount": 1,
  "LastEvaluatedKey": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "7"
    }
  }
}

D, [2020-05-28T20:57:28.750503 #35684] DEBUG -- : Query

Request:

{
  "ConsistentRead": false,
  "ScanIndexForward": true,
  "Limit": 1,
  "TableName": "dynamoid_tests_documents_1590688648_522",
  "KeyConditions": {
    "name": {
      "ComparisonOperator": "EQ",
      "AttributeValueList": [
        {
          "S": "Alice"
        }
      ]
    }
  },
  "QueryFilter": {
    "amount": {
      "ComparisonOperator": "EQ",
      "AttributeValueList": [
        {
          "N": "3"
        }
      ]
    }
  },
  "ExclusiveStartKey": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "7.0"
    }
  }
}

Response:

{
  "Items": [
    {
      "name": {
        "S": "Alice"
      },
      "created_at": {
        "N": "1590688648.693225"
      },
      "amount": {
        "N": "3"
      },
      "updated_at": {
        "N": "1590688648.693261"
      },
      "age": {
        "N": "8"
      }
    }
  ],
  "Count": 1,
  "ScannedCount": 1,
  "LastEvaluatedKey": {
    "name": {
      "S": "Alice"
    },
    "age": {
      "N": "8"
    }
  }
}
```
</details>

Please note there are 3 logged Query requests (`DEBUG -- : Query`) and only the last one returns an item. Despite the fact it returns correct result it isn't as efficient as we would expect.

I suppose there can be situations with extra large items and loading all of them will be longer than sending multiple requests with empty results.

So I would consider such optimization only for queries with hash key/range key conditions.

mrkamel:
Interesting, thx for pointing this out.

Changed the code to only apply the record limit if no non-key conditions are present, i.e. for queries like:

```ruby
model.first
model.where(name: 'Bob').first
model.where(name: 'Bob', age: 1).first
```

and no record limit will be applied for queries like:

```ruby
model.where(age: 1).first # hash key missing
model.where(name: 'Bob', city: 'New York').first # city is not a key
# ...
```

Actually, i propose to extend it a bit, such that `record_limit(1000)` is applied when non-keys are included. This would have the benefit that one would no longer run into the issue to accidently fetch **everything** or a large subset to just get the first item and at the same time, avoids this 1-step issue you mentioned. What do you think?

lib/dynamoid/criteria/chain.rb
```diff
@@ -176,6 +176,27 @@ def count
...
+      #
+      # If used without criteria it just returns the first item of some
+      # arbitrary order.
+      #
+      #   Post.first
+      #
+      # @return [Model|nil]
+      def first
+        if @key_fields_detector.non_key_present?
+          all.to_a.first
```
andrykonchin, line 194: Method `all` returns lazy enumerator to iterate over paginated result. So it makes sense not to materialize the whole collection with calling `to_a` method here.

lib/dynamoid/criteria/chain.rb
```diff
@@ -176,6 +176,27 @@ def count
...
+      # when no non-key conditions are present.
+      #
+      # If used without criteria it just returns the first item of some
+      # arbitrary order.
+      #
+      #   Post.first
+      #
+      # @return [Model|nil]
+      def first
+        if @key_fields_detector.non_key_present?
```
andrykonchin, line 193: I believe this condition should be more strict - only Query operations w/o non key conditions should be optimized.

Condition on a hash key should be equality like `{ id: 1 }`. Not-equality condition for a hash key means the Scan operation will be used instead of the Query. And Scan operations combined with `record_limit` method call leads to the issue with multiple requests.

andrykonchin:
> Actually, i propose to extend it a bit, such that record_limit(1000) is applied when non-keys are included. This would have the benefit that one would no longer run into the issue to accidently fetch everything or a large subset to just get the first item and at the same time, avoids this 1-step issue you mentioned. What do you think?

Hmm. Yes, it sounds reasonable. But it doesn't solve the issue but just postpones it. And other users will pay for it even if they don't have such issue.

DynamoDB returns Query/Scan result paginated by 1Mb chunks ([documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html)) so nobody will read the whole table just to get the first item. I still don't know any good enough solution for this issue.

But I think there is still room for improvement. If there are no any conditions at all e.g.

```ruby
Post.first
```

we can limit the read items with `scan_limit(1)` and not to load the whole first page of a table.

lib/dynamoid/criteria/chain.rb
mrkamel: delegating to `super` now

lib/dynamoid/criteria/chain.rb
mrkamel: actually, already the case. Maybe naming of `KeyFieldsDetector#non_key_present?` in  confused you. But naming of `KeyFieldsDetector#key_present?` doesn't tell me that equality operation is checked use either. So, added a spec for it.

mrkamel:
> we can limit the read items with scan_limit(1) and not to load the whole first page of a table.

:+1: done

lib/dynamoid/criteria/chain.rb
andrykonchin: Agree 👍

andrykonchin:
Good job 👍 .

Could you please squash commits? There are at least two optimizations and it's up to you how to commit them.

There is a minor issue with specs but they are OK and the PR may be merged as is. Some new specs are isolated and don't send actual requests to DynamoDB. They use method expectations and don't check actual result e.g.

```ruby
expect(chain).to receive(:record_limit).with(1)
```

There are very few isolated tests in the gem and I would prefer not isolated specs. To check whether some method (e.g. `record_limit`) was called and still don't mock it I would use `and_call_original` RSpec method:

```ruby
chain = Dynamoid::Criteria::Chain.new(model)
expect(chain).to receive(:raw_pages_via_query).and_call_original
expect(chain.where(name: 'Bob', age: 10).all).to contain_exactly(customer1)
```

mrkamel:
... squashed them, such that 2 commits remain, one for the record_limit optimization, one for the scan_limit optimization.

In addition, i changed the specs to add expectations on the return value.

andrykonchin:
Thank you for this improvement 👍

PR #445, opened 2020-06-04
Support n parameter in Criteria::Chain#first

mrkamel:
I recognized we forgot about the optional n parameter in Criteria::Chain#first, such that this PR adds support for it by applying the param to scan_limit respectively record_limit.


andrykonchin:
Makes sense. Thank you.

PR #447, opened 2020-06-12
Fix Ruby 2.7 keyword argument deprecation warnings

geek4good:
Serveral methods had been written in a way that calling them resulted in the
following deprecation warning when using Dynamoid with Ruby 2.7:

warning: Using the last argument as keyword parameters is deprecated; maybe **
should be added to the call

This change rectifies this by explicitly using keyword arguments instead of
implicitly relying on Ruby to do the right thing.

### DIFF
--- a/lib/dynamoid/adapter.rb
+++ b/lib/dynamoid/adapter.rb
@@ -155,9 +155,15 @@ module Dynamoid
     #
     # @since 0.2.0
     def method_missing(method, *args, &block)
-      return benchmark(method, *args) { adapter.send(method, *args, &block) } if adapter.respond_to?(method)
+      super unless adapter.respond_to?(method)
 
-      super
+      if args.last.is_a?(Hash)
+        return benchmark(method, *args[0..-2], **args.last) do
+          adapter.send(method, *args[0..-2], **args.last, &block)
+        end
+      end
+
+      benchmark(method, *args) { adapter.send(method, *args, &block) }
     end
 
     # Query the DynamoDB table. This employs DynamoDB's indexes so is generally faster than scanning, but is
--- a/lib/dynamoid/loadable.rb
+++ b/lib/dynamoid/loadable.rb
@@ -23,7 +23,7 @@ module Dynamoid
         options[:range_key] = range_value
       end
 
-      self.attributes = self.class.find(hash_key, options).attributes
+      self.attributes = self.class.find(hash_key, **options).attributes
       @associations.values.each(&:reset)
       self
     end
--- a/lib/dynamoid/persistence/update_fields.rb
+++ b/lib/dynamoid/persistence/update_fields.rb
@@ -4,7 +4,7 @@ module Dynamoid
   module Persistence
     class UpdateFields
       def self.call(*args)
-        new(*args).call
+        new(args.first, **args.last).call
       end
 
       def initialize(model_class, partition_key:, sort_key:, attributes:, conditions:)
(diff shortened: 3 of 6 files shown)


lib/dynamoid/loadable.rb
```diff
@@ -23,7 +23,7 @@ def reload
         options[:range_key] = range_value
       end
 
-      self.attributes = self.class.find(hash_key, options).attributes
+      self.attributes = self.class.find(hash_key, **options).attributes
```
andrykonchin, line 26: 👍

lib/dynamoid/persistence/update_fields.rb
```diff
@@ -4,7 +4,7 @@ module Dynamoid
   module Persistence
     class UpdateFields
       def self.call(*args)
-        new(*args).call
+        new(args.first, **args.last).call
```
andrykonchin, line 7: Doesn't  `def self.call(*args, **kwargs)` look better?

andrykonchin:
Thank you for addressing this issue.

Despite I don't have strong opinion how to solve the issue generally there are several options right now. There is a comprehensive enough post in @eregon's blog post [The Delegation Challenge of Ruby 2.7](https://eregon.me/blog/2019/11/10/the-delegation-challenge-of-ruby27.html)

lib/dynamoid/adapter.rb
```diff
@@ -155,9 +155,15 @@ def delete_table(table_name, options = {})
...
+      super unless adapter.respond_to?(method)
 
-      super
+      if args.last.is_a?(Hash)
+        return benchmark(method, *args[0..-2], **args.last) do
+          adapter.send(method, *args[0..-2], **args.last, &block)
+        end
+      end
+
+      benchmark(method, *args) { adapter.send(method, *args, &block) }
```
andrykonchin, line 166: Again, what do you think about changing method signature?

```
def method_missing(method, *args, **kwargs, &block)
```

eregon:
Only `ruby2_keywords` is correct for delegation if you want to support Ruby 2.7:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/
`*args, **kwargs`-delegation isn't fully correct on 2.7, unfortunately.

lib/dynamoid/adapter.rb
geek4good: Thank you for the review and feedback. Looks as is using `ruby2_keywords` is the way to go with [defining a dummy version of it for Ruby 2.6 and prior](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) (thanks @eregon):

```ruby
def ruby2_keywords(*)
end if RUBY_VERSION < "2.7"
```

andrykonchin: Yeah. Looks like it's the easiest way.

But comparing string versions `RUBY_VERSION < "2.7"` isn't correct for minor versions like '2.10' and greater. I like the solution from https://github.com/ruby/ruby2_keywords:

```ruby
class Module
  unless respond_to?(:ruby2_keywords, true)
    private
    def ruby2_keywords(name, *)
      # nil
    end
  end
end
```

But I would like to avoid monkey matching global classes.

andrykonchin:
Fixed in https://github.com/Dynamoid/dynamoid/pull/482

Issue #449, opened 2020-06-15
Is it possible to not using null-logger gem, because of its license


yakjuly:
Hi, I am trying to use dynamoid gem in a commericial project. 

But the company not allowing to use open source libraries that under GPL and LGPL license.

dynamoid is under MIT license, but it is using null-logger and null-logger is under GNU LGPL license. Kind of blocker.

I am thinking since dynamoid doesn't use NullLogger in many places, is it possible to replace it with a customized logger. After that dynamoid can be used in more places.

andrykonchin:
Hmmm. Good point. Never thought about it. A PR with replacing `NullLogger` is welcome.

PR #450, opened 2020-06-16
Replace NullLogger with Logger.new(nil)

yakjuly:
https://github.com/Dynamoid/dynamoid/issues/449

NullLogger is GNU LGPL v3 license which may limit commercial use.
To avoid that, use `Logger.new(nil)` instead.


andrykonchin:
Thank you

PR #461, opened 2020-07-16
Allow attribute deletion without a value

kirsis:
Currently it seems that once an attribute is created, it can only be set to null, or - if it is a set/array - a value can be deleted from it.

With this change, the attribute appears to be deleteable entirely.


andrykonchin:
Looks good. Thank you for this improvement.

PR #463, opened 2020-07-21
Raising UnknownAttribute when adding an attribute not in the model

### LINKED ISSUE #462: NoMethodError when trying to add an attribute that isn't on the model
### Bug description
When using `#update_attributes` to update a field not present in the model, the method completes but trying to save the object raises an error when trying to dump it. The root cause is on [lib/dynamoid/dumping.rb#L6-L41](https://github.com/Dynamoid/dynamoid/blob/master/lib/dynamoid/dumping.rb#L6-L41), because if the attribute is not present in `attributes_options`, the `options` argument passed to `find_dumper` will be `nil`.

While I understand that it may be on purpose to not allow unknown fields on the model, returning a detailed error message will simplify the debugging and result in a better developer experience for Dynamoid's users. I'm happy to prepare a PR to fix the issue, but would like to confirm what should be the expected behavior here: should the error be raised on `#update_attributes`, or would you be fine if the validation and error raising is done on the `Dumping` class directly?

### Steps to reproduce
* **Dynamoid version:** _master_
* **Ruby version:** 2.6.3

I've prepared a small test spec to easily reproduce the error:

```ruby
  describe 'Field that is not declared in the model' do
    it 'raises a NoMethodError' do
      obj = new_class.create
      obj.update_attributes(unknown_field: 'Some value')
    end
  end
```

This is the output:

```ruby
     Failure/Error:
       dumper_class = case options[:type]
                      when :string     then StringDumper
                      when :integer    then IntegerDumper
                      when :number     then NumberDumper
                      when :set        then SetDumper
                      when :array      then ArrayDumper
                      when :map        then MapDumper
                      when :datetime   then DateTimeDumper
                      when :date       then DateDumper
                      when :serialized then SerializedDumper

     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/dynamoid/dumping.rb:27:in `find_dumper'
     # ./lib/dynamoid/dumping.rb:17:in `dump_field'
     # ./lib/dynamoid/dumping.rb:9:in `block (2 levels) in dump_attributes'
     # ./lib/dynamoid/dumping.rb:8:in `each'
     # ./lib/dynamoid/dumping.rb:8:in `block in dump_attributes'
     # ./lib/dynamoid/dumping.rb:7:in `tap'
     # ./lib/dynamoid/dumping.rb:7:in `dump_attributes'
     # ./lib/dynamoid/persistence/save.rb:23:in `call'
     # ./lib/dynamoid/persistence/save.rb:8:in `call'
     # ./lib/dynamoid/persistence.rb:483:in `block in save'
     # ./lib/dynamoid/persistence.rb:482:in `save'
     # ./lib/dynamoid/dirty.rb:50:in `save'
     # ./lib/dynamoid/validations.rb:19:in `save'
     # ./lib/dynamoid/identity_map.rb:68:in `save'
     # ./lib/dynamoid/persistence.rb:499:in `update_attributes'
     # ./spec/dynamoid/dumping_spec.rb:1515:in `block (3 levels) in <top (required)>'
     # ./spec/support/unregister_declared_classes.rb:6:in `block (2 levels) in <top (required)>'
```

AlexGascon:
Fixes #462 

### Changes
Modified the method `write_attribute` to raise an `UnknownAttribute` error if the user tries to write a field that hasn't been defined on the model class

Created a new module, `Dynamoid::Persistence::UpdateValidations`, with a method to validate if a specific attribute is part of a model class

Adding tests to validate the new behavior in:
- `.upsert`
- `.update_fields`
- `.update`
- `#update`
- `#update_attribute`
- `#update_attributes`

### DIFF
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -277,6 +277,9 @@ module Dynamoid
       # +update_fields+ uses the +UpdateItem+ operation so it saves changes and
       # loads an updated document back with one HTTP request.
       #
+      # Raises a +Dynamoid::Errors::UnknownAttribute+ exception if any of the
+      # attributes is not on the model
+      #
       # @param hash_key_value [Scalar value] hash key
       # @param range_key_value [Scalar value] range key (optional)
       # @param attrs [Hash]
@@ -324,6 +327,9 @@ module Dynamoid
       #
       # +upsert+ uses the +UpdateItem+ operation so it saves changes and loads
       # an updated document back with one HTTP request.
+      #
+      # Raises a +Dynamoid::Errors::UnknownAttribute+ exception if any of the
+      # attributes is not on the model
       #
       # @param hash_key_value [Scalar value] hash key
       # @param range_key_value [Scalar value] range key (optional)
@@ -492,14 +498,15 @@ module Dynamoid
     #
     #   user.update_attributes(age: 27, last_name: 'Tylor')
     #
+    # Raises a +Dynamoid::Errors::UnknownAttribute+ exception if any of the
+    # attributes is not on the model
+    #
     # @param attributes [Hash] a hash of attributes to update
     # @return [true|false] Whether updating successful or not
     # @since 0.2.0
     def update_attributes(attributes)
       attributes.each { |attribute, value| write_attribute(attribute, value) }
       save
-    rescue Dynamoid::Errors::UnknownAttribute
-      false
     end
 
     # Update multiple attributes at once, saving the object once the updates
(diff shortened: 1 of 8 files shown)


lib/dynamoid/persistence.rb
```diff
@@ -497,6 +498,8 @@ def save(options = {})
     def update_attributes(attributes)
       attributes.each { |attribute, value| write_attribute(attribute, value) }
       save
+    rescue Dynamoid::Errors::UnknownAttribute
+      false
```
andrykonchin, line 502: I think it makes sense to follow Rails convention and raise `UnknownAttribute` exception anyway even in methods without bang.

andrykonchin:
Looks pretty good 👍 . There is only one issue to discuss.

lib/dynamoid/persistence.rb
AlexGascon: Sounds good to me, I'll make the change

andrykonchin:
Good job. Thank you for the contribution.

PR #468, opened 2020-08-24
Add write transaction

btree1970:
Add transaction write feature to DynamoId.

- Currently only works for adding multiple write operations

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
@@ -180,6 +180,27 @@ module Dynamoid
         end
       end
 
+      def transact_write_items(table_name, models, _options = {})
+        out_put = process(table_name, models)
+        begin
+          response = client.transact_write_items({
+
+                                                   transact_items: out_put,
+            return_consumed_capacity: 'TOTAL',
+            return_item_collection_metrics: 'SIZE'
+                                                 })
+        rescue StandardError => e
+          puts e
+        end
+      end
+
+      def process(table_name, objects)
+        objects.map do |m|
+          key, attrib = m.flatten
+          { key => { item: sanitize_item(attrib), table_name: table_name } }
+        end
+      end
+
       # Get many items at once from DynamoDB. More efficient than getting each item individually.
       #
       # If optional block is passed `nil` will be returned and the block will be called for each read batch of items,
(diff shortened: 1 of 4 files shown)


andrykonchin:
Looks pretty good 👍 .

The only issue I see is that `transact` method works just for the **one** model/table but DynamoDB's transaction can involve different tables in one call. I would like to have only **one** flexible API for write transactions. That's why I would like `transact` to work with several tables.

The only reason why transactions aren't added yet is that I am not sure what API should be.

I have following thoughts:
* API should be high-level and operate model classes instead of table names
* operations are hight-level and names match already existing methods like `update_attributes`/`update_fields`/`upsert`/...
* callbacks should be called, at least some of them (e.g. for creation/deletion)

I considered following interface:

```ruby
# Dyhamoid::Document.transaction or User.transaction

Dynamoid::Document.transaction do |txn|
  txn.create(User, attributes)
  txn.update(User, attributes)
  txn.update_fields(User, attributes, conditions)
  txn.upsert(User, attributes)
  txn.delete(User, id)
end

So please let me know what do you think about it.

btree1970:
> Looks pretty good 👍 .
> 
> The only issue I see is that `transact` method works just for the **one** model/table but DynamoDB's transaction can involve **different** tables in one call. I would like to have only **one** flexible API for write transactions. That's why I would like `transact` to work with several tables.
> 
> The only reason why transactions aren't added yet is that I am not sure what API should be.
> 
> I have following thoughts:
> 
> * API should be high-level and operate model classes instead of table names
> * operations are hight-level and names match already existing methods like `update_attributes`/`update_fields`/`upsert`/...
> * callbacks should be called, at least some of them (e.g. for creation/deletion)
> 
> I considered following interface:
> 
> ```ruby
> # Dyhamoid::Document.transaction or User.transaction
> 
> Dynamoid::Document.transaction do |txn|
>   txn.create(User, attributes)
>   txn.update(User, attributes)
>   txn.update_fields(User, attributes, conditions)
>   txn.upsert(User, attributes)
>   txn.delete(User, id)
> end
> 
> So please let me know what do you think about it.
> ```

Hey thanks for the response. I admit i did this PR with a specific business case in mind. I like your suggestion for the new api design. Will work on it over the next few days.

dpsk:
@andrykonchin what's the status on transact write support, is it planned? any help needed?

andrykonchin:
Transactions aren't supported right now. The main issue is that it isn't clear to me how the transactions API should look. Any opinion/proposition is welcomed.

ckhsponge:
Have there been any more thoughts on this? Below is an alternate API idea using an array of arrays. An array of hashes could work too. Personally I prefer single table usage of DynamoDB so don't have a need to span tables but I see why it should be supported.  
`
Dynamoid::Document.transaction([
  [User, :create, attributes],
  [User, :update, attributes],
  [User, :update_fields, attributes, conditions],
  [User, :upsert, attributes],
  [User, :delete, id]
])
`

PR #470, opened 2020-09-24
Minor consistency fix

rromanchuk:
This example fails in real life, scheme is required
> ArgumentError (expected :endpoint to be a HTTP or HTTPS endpoint)


andrykonchin:
Thank you for the fix 👍

PR #473, opened 2020-11-22
Update readme: fix "Dynamid" -> "Dynamoid" in ch. "Type casting"

Rulikkk:


andrykonchin:
Thank you for the fix!

PR #476, opened 2020-12-08
Adding #with_index to chain to force allow the user to force an index

bmalinconico:
This PR adds the ability to force an index selection when constructing a query chain. 

There are queries, which would fit an existing index but not return results in the desired order. The example from the method docs:

```ruby
class Comment
  include Dynamoid::Document

  hash_key :post_id
  range_key :author_id

  field :post_date, :datetime

  local_secondary_index name: :time_sorted_comments, range_key: post_date, projected_attributes: :all
end


 Comment.where(post_id: id).with_index(:time_sorted_comments).scan_index_forward(false)
```

The current behavior of Dynamoid would only select this index if I also filtered by the secondary index. This PR allows me to not add a null-op filter, and just hint at the index.

There are open questions in the PR:

1) What do we want to do when the named index is not found
2) What do we do when the user materializes a query which cannot be applied to the index? (Isn't performing equality on the index hash key)

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -110,7 +110,7 @@ module Dynamoid
         query.update(args.symbolize_keys)
 
         # we should re-initialize keys detector every time we change query
-        @key_fields_detector = KeyFieldsDetector.new(@query, @source)
+        @key_fields_detector = KeyFieldsDetector.new(@query, @source, forced_index_name: @forced_index_name)
 
         self
       end
@@ -355,6 +355,36 @@ module Dynamoid
       # @return [Dynamoid::Criteria::Chain]
       def scan_index_forward(scan_index_forward)
         @scan_index_forward = scan_index_forward
+        self
+      end
+
+      # Force the index name to use for queries.
+      #
+      # By default allows the library to select the most appropriate index.
+      # Sometimes you have more than one index which will fulfill your query's
+      # needs. When this case occurs you may want to force an order. This occurs
+      # when you are searching by hash key, but not specifying a range key.
+      #
+      #  class Comment
+      #    include Dynamoid::Document
+      #
+      #    table key: :post_id
+      #    range_key :author_id
+      #
+      #    field :post_date, :datetime
+      #
+      #    global_secondary_index name: :time_sorted_comments, hash_key: :post_id, range_key: post_date, projected_attributes: :all
+      #  end
+      #
+      #
+      #   Comment.where(post_id: id).with_index(:time_sorted_comments).scan_index_forward(false)
+      #
+      # @return [Dynamoid::Criteria::Chain]
+      def with_index(index_name)
+        raise Dynamoid::Errors::InvalidIndex, "Unknown index #{index_name}" unless @source.find_index_by_name(index_name)
+
+        @forced_index_name = index_name
+        @key_fields_detector = KeyFieldsDetector.new(@query, @source, forced_index_name: index_name)
         self
       end
 
--- a/lib/dynamoid/criteria/key_fields_detector.rb
+++ b/lib/dynamoid/criteria/key_fields_detector.rb
@@ -24,10 +24,11 @@ module Dynamoid
         end
       end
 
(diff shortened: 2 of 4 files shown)


lib/dynamoid/criteria/key_fields_detector.rb
```diff
@@ -133,6 +136,31 @@ def match_global_secondary_index
           }
         end
       end
+
+      def match_forced_index
+        idx = possilbe_local_secondary_index || possible_global_secondary_index
+
+        # TODO: Handle named index not found
+        # TODO: Check equality condition exists on requested index hash key
```
andrykonchin, line 144: 1. I think we should raise exception when index name isn't found, e.g. `Dynamoid::Errors::InvalidIndex`.
2. It's OK to check name here but I would say it makes sense to check it in `Chain#with_index` method.
3. Using forced index name without equality condition may seem useless (because it means Scan operation instead of Query) but it's still a valid case and it could be reasonable for instance when GSI is used and not all attributes are projected - this way the operation will be cheaper. So I don't see any reason to check equality condition at all here.

lib/dynamoid/criteria/key_fields_detector.rb
```diff
@@ -24,10 +24,11 @@ def contain_with_eq_operator?(field_name)
         end
       end
 
-      def initialize(query, source)
+      def initialize(query, source, forced_index_name)
```
andrykonchin, line 27: **suggestion:** I would prefer using keyword argument for optional parameters
```suggestion
      def initialize(query, source, forced_index_name: nil)
```

lib/dynamoid/criteria/chain.rb
```diff
@@ -358,6 +359,34 @@ def scan_index_forward(scan_index_forward)
...
+      #    global_secondary_index name: :time_sorted_comments, hash_key: :post_id, range_key: post_date, projected_attributes: :all
+      #  end
+      #
+      #
+      #   Comment.where(post_id: id).with_index(:time_sorted_comments).scan_index_forward(false)
+      #
+      # @return [Dynamoid::Criteria::Chain]
+      def with_index(index_name)
+        @forced_index_name = index_name
+        @key_fields_detector = KeyFieldsDetector.new(@query, @source, index_name)
```
andrykonchin, line 386: Well done. `KeyFieldsDetector` is an ideal place for this logic.

lib/dynamoid/criteria/chain.rb
```diff
@@ -358,6 +359,34 @@ def scan_index_forward(scan_index_forward)
...
+      #
+      # By default allows the library to select the most appropriate index.
+      # Sometimes you have more than one index which will fulfill your query's
+      # needs. When this case occurs you may want to force an order. This occurs
+      # when you are searching by hash key, but not specifying a range key.
+      #
+      #  class Comment
+      #    include Dynamoid::Document
+      #
+      #    hash_key :post_id
```
andrykonchin, line 372: **monir:**
```suggestion
      #    table key: :post_id
```

andrykonchin (commented):
I've been considered adding this method for a long time but hesitated because wasn't sure whether anybody needs it.

Well done 👍

lib/dynamoid/criteria/key_fields_detector.rb
bmalinconico: I've updated all instances of the class usage to pass a value, albeit null. I wasn't sure if there would be a need to take in on initialization at some point in the future. This way all constructor calls are the same three instance vars.

I can revert that change to make it an optional keyword param if you prefer that to passing a known null value.

andrykonchin:
Could you please :
- remove TODOs
- raise exception when specified non-existing index name
- squash working commits
?

bmalinconico:
@andrykonchin Yep working on this now, paid work got in the way yesterday

bmalinconico:
@andrykonchin Pushed the updates. I moved the index find by name method into the model's class methods since it was used in two places.

andrykonchin:
Good job 👍

Issue #467, opened 2020-08-18
Parent association record of object that :belongs_to another object not deleted on object delete/destroy


mitchellharper12:
Rails version: 6.0.3.2
Dynamoid version: 3.6.0

Expected behavior:

When I delete or destroy an object of type Child that belongs to another object Parent, and Parent has many Child objects, the foreign key on the parent association record is deleted

Actual behavior:

The foreign key is not deleted, and trying to resolve elements in e.g. Parent.find(id).users results in an error like the following

```
Dynamoid::Errors::RecordNotFound (Couldn't find all Users with primary keys [d203aa5e-78ab-4763-ad7f-2636311a2e42] (found 0 results, but was looking for 1))
```

Will try and get to the root of the issue myself and add a PR if I can.

andrykonchin:
Nice catch.

Yeah, looks like associations aren't updated when document is deleted/destroyed. Any PR is welcome 👍 .

PR #484, opened 2021-01-11
Update associations on object delete

kkan:
Resolves issue #467

### DIFF
--- a/lib/dynamoid/associations/association.rb
+++ b/lib/dynamoid/associations/association.rb
@@ -58,6 +58,12 @@ module Dynamoid
         :set
       end
 
+      def disassociate_source
+        Array(target).each do |target_entry|
+          target_entry.send(target_association).disassociate(source.hash_key) if target_entry && target_association
+        end
+      end
+
       private
 
       # The target class name, either inferred through the association's name or specified in options.
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -801,9 +801,20 @@ module Dynamoid
 
       @destroyed = true
 
-      Dynamoid.adapter.delete(self.class.table_name, hash_key, options)
+      Dynamoid.adapter.delete(self.class.table_name, hash_key, options).tap { update_asociations }
     rescue Dynamoid::Errors::ConditionalCheckFailedException
       raise Dynamoid::Errors::StaleObjectError.new(self, 'delete')
+    end
+
+    private
+
+    def update_asociations
+      self.class.associations.each do |name, options|
+        begin
+          send(name).disassociate_source
+        rescue Aws::DynamoDB::Errors::ResourceNotFoundException
+        end
+      end
     end
   end
 end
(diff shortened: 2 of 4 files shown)


andrykonchin:
Tank you for the fix. I need some time to check it.

lib/dynamoid/persistence.rb
```diff
@@ -801,9 +801,20 @@ def delete
 
       @destroyed = true
 
-      Dynamoid.adapter.delete(self.class.table_name, hash_key, options)
+      Dynamoid.adapter.delete(self.class.table_name, hash_key, options).tap { update_asociations }
```
andrykonchin, line 804: **minor:** TBH I would avoid introducing new method (`update_asociations`) which can conflict with user defined methods in model and would inline it into the `delete` method.

spec/dynamoid/persistence_spec.rb
```diff
@@ -2597,6 +2597,38 @@ def log_message
...
+        expect do
+          user.delete
+        end.to change { Magazine.find(magazine.title).owner.target }.from(user).to(nil)
+      end
+
+      it 'updates has_one association' do
+        expect do
+          user.delete
+        end.to change { Subscription.find(monthly.id).customer.target }.from(user).to(nil)
+      end
```
andrykonchin, line 2624: The only way this fix could be improved is to ignore `RecordNotFound` errors when association is loaded and there are some broken foreign keys (what is described in the original issue https://github.com/Dynamoid/dynamoid/issues/467) . This way we would solved the problem for existing corrupted references.

But it's out of scope of this PR and it's up to you.

lib/dynamoid/associations/association.rb
```diff
@@ -58,6 +58,12 @@ def declaration_field_type
         :set
       end
 
+      def disassociate_source
+        Array(target).each do |target_entry|
+          target_entry.send(target_association).disassociate(source.hash_key) if target_entry && target_association
+        end
+      end
```
andrykonchin, line 65: **minor:** As far as it's a new public method - it could be covered with unit tests as well.

andrykonchin (commented):
Looks good 👍 

I have only one concern but it doesn't prevent merging this fix.

There are actually two issues
- bug in the internal Dynamoid's mechanism (in context of `belongs_to` and `has_and_belongs_to_many` associations) when cached ids for `belongs_to` and `has_and_belongs_to_many` should be updated and id of deleted item should be removed
- referential integrity (in context of `has_one` and `has_many` associations) - when parent model is deleted - reference in child model becomes invalid

This fix solves both problems. But by Rails' convention there is a difference between `#delete` and `#destroy` model methods - `#delete` just removes a record from a table but `#destroy` loads a model from a database, runs proper callback (as far as I remember Rails doesn't run callbacks at deleting) and ensures referential integrity - deletes/destroy child model or just nullify foreign key.

It's important for Dynamoid to follow existing Rails conventions and it would be great to have the same difference between `#delete` and `#destroy` model methods as well. This way `#delete` should handle only `belongs_to` and `has_and_belongs_to_many` associations but `#destroy` should also ensures referential integrity and update `has_one` and `has_many` associations as well (nullify foreign keys).

PR #492, opened 2021-02-14
Support Date index

yoshida-eth0:
Hi!
This PR supports `:date` type index.

`Dynamoid::Errors::InvalidIndex` occurs when `:date` type field is specified for hash_key or range_key.

```
Validation failed: Range key Index :range_key is not a valid key type (Dynamoid::Errors::InvalidIndex)
```


andrykonchin:
Thank you for the fix 👍 .

PR #496, opened 2021-03-18
add http_proxy and create_table_enabled config options

HLH15:
Add config options for
* http_proxy: DynamoDB http_proxy
* create_table_enabled: to avoid creating tables in DynamoDB directly through Dynamoid.


andrykonchin:
Could you please fix the failing specs?

HLH15:
> Could you please fix the failing specs? Current master passes specs successfuly.

@andrykonchin This is fixed now!

andrykonchin:
It looks good to me. But I am wondering what a use case for disabling table creation.

I assume it makes sense to prevent lazy table creation at model saving. In this case, it can be checked right here in the [#save](https://github.com/Dynamoid/dynamoid/blob/v3.7.0/lib/dynamoid/persistence.rb#L478) method

HLH15:
> It looks good to me. But I am wondering what a use case for disabling table creation.
> 
> I assume it makes sense to prevent lazy table creation at model saving. In this case, it can be checked right here in the [#save](https://github.com/Dynamoid/dynamoid/blob/v3.7.0/lib/dynamoid/persistence.rb#L478) method

In our use case, we have a different pipeline to manage all AWS resources (Terraform). So we'd like to prevent creating new tables in all use cases.

andrykonchin:
I see. I would prefer to change the `#save` method instead of the `#create_table`. Also, I think we need to add tests on `#save` and `#create` and check that table isn't created implicitly.

PR #498, opened 2021-03-29
Fix JRuby CI failure

dmitry-ilyashevich:


andrykonchin:
Thank you for the fix 👍

PR #503, opened 2021-04-30
Fix scan with index

bmalinconico:
I found a bug when forcing and index and not filtering by hash key. This fixes that and pushes the logic down into the adapter.

I'll be back to this on monday, please let me know if you would like to try and satisfy code climate's complexity concerns.


andrykonchin:
Looks pretty good 👍 . Thank you.

PR #512, opened 2021-06-14
Update #delete_all to skip callbacks

jaybee-heon:
As a rails developer, I expected that method 'delete_all' to bypass callbacks.
Is it good or bad to skip callbacks also in dynamoid?

Please give me any feedback!
Thank you in advance.

### DIFF
--- a/lib/dynamoid/associations/many_association.rb
+++ b/lib/dynamoid/associations/many_association.rb
@@ -177,7 +177,8 @@ module Dynamoid
       # @since 0.2.0
       def delete_all
         objs = target
-        source.update_attribute(source_attribute, nil)
+        source.write_attribute(source_attribute, nil)
+        source.save(skip_callbacks: true)
         objs.each(&:delete)
       end
 
--- a/lib/dynamoid/persistence.rb
+++ b/lib/dynamoid/persistence.rb
@@ -478,16 +478,25 @@ module Dynamoid
       self.class.create_table(sync: true)
 
       @_touch_record = options[:touch]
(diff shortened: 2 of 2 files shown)


andrykonchin:
It's a quite welcome improvement 🎆 .

Dynamoid tries to be as compatible with Rails' conventions and API as possible (but associations were implemented in a too different way unfortunatelly).

lib/dynamoid/associations/many_association.rb
```diff
@@ -177,7 +177,8 @@ def destroy_all
       # @since 0.2.0
       def delete_all
         objs = target
-        source.update_attribute(source_attribute, nil)
+        source.write_attribute(source_attribute, nil)
+        source.save(skip_callbacks: true)
```
andrykonchin, line 181: Looks good. But it can be improved with `#update` or `update_fields` methods. A least in this case we don't need to add the `skip_callbacks` method. As far as I know, there is no similar option (for the `#save` method) in Rails.

jaybee-heon:
> It's a quite welcome improvement 🎆 .
> 
> Dynamoid tries to be as compatible with Rails' conventions and API as possible (but associations were implemented in a too different way unfortunatelly).

Oh I see! Thank you for your comment!
I'll keep that in mind.
Hope I could contribute this project 🙏

lib/dynamoid/associations/many_association.rb
jaybee-heon: Thank you for your feedback!
I'll comeback with better changes.

PR #525, opened 2021-12-21
Don't call attribute_will_change! if the new values equals the old value

a5-stable:
Don't call attribute_will_change! if the new values equals the old value

### DIFF
--- a/lib/dynamoid/fields.rb
+++ b/lib/dynamoid/fields.rb
@@ -294,7 +294,7 @@ module Dynamoid
     # @since 0.2.0
     def write_attribute(name, value)
       name = name.to_sym
-      old_value = __send__(name)
+      old_value = read_attribute(name)
 
       unless attribute_is_present_on_model?(name)
         raise Dynamoid::Errors::UnknownAttribute.new("Attribute #{name} is not part of the model")
@@ -307,9 +307,9 @@ module Dynamoid
       @attributes_before_type_cast[name] = value
 
       value_casted = TypeCasting.cast_field(value, self.class.attributes[name])
-      attributes[name] = value_casted
-
       attribute_will_change!(name) if old_value != value_casted # Dirty API
+
+      attributes[name] = value_casted
     end
     alias []= write_attribute
 
--- a/lib/dynamoid/fields.rb
+++ b/lib/dynamoid/fields.rb
@@ -369,7 +369,7 @@ module Dynamoid
     # @since 0.2.0
     def set_updated_at
       # @_touch_record=false means explicit disabling
-      if self.class.timestamps_enabled? && !updated_at_changed? && @_touch_record != false
+      if self.class.timestamps_enabled? && !changed? && @_touch_record != false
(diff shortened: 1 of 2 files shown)


lib/dynamoid/fields.rb
```diff
@@ -294,6 +294,7 @@ def generated_methods
     # @since 0.2.0
     def write_attribute(name, value)
       name = name.to_sym
+      old_value = __send__(name)
```
andrykonchin, line 297: `read_attribute` looks better than `__send__`

lib/dynamoid/fields.rb
```diff
@@ -303,12 +304,12 @@ def write_attribute(name, value)
...
       end
 
-      attribute_will_change!(name) # Dirty API
-
       @attributes_before_type_cast[name] = value
 
       value_casted = TypeCasting.cast_field(value, self.class.attributes[name])
       attributes[name] = value_casted
+
+      attribute_will_change!(name) if old_value != value_casted # Dirty API
```
andrykonchin, line 312: `attribute_will_change!` should be called before assigning a new value because it saves the previous old one.

andrykonchin (commented):
It will be great to have this bug covered with specs. The proper file I think is https://github.com/Dynamoid/dynamoid/blob/master/spec/dynamoid/dirty_spec.rb

lib/dynamoid/fields.rb
a5-stable: Thanks, I fixed it.

lib/dynamoid/fields.rb
a5-stable: I fixed it!

lib/dynamoid/fields.rb
```diff
@@ -368,7 +369,7 @@ def set_created_at
     # @since 0.2.0
     def set_updated_at
       # @_touch_record=false means explicit disabling
-      if self.class.timestamps_enabled? && !updated_at_changed? && @_touch_record != false
+      if self.class.timestamps_enabled? && !changed? && @_touch_record != false
```
a5-stable, line 372: !updated_at_changed? -> !changed?
because ActiveRecord modifies updated_at field only when changed? is true.

andrykonchin: The idea here is that we don't override `updated_at` provided by user:

```ruby
model.update_attributes(updated_at: Time.now - 1.month)
```

a5-stable: Thanks for your quick feedback. Absolutely.

now I changed the condition like this;
`if self.class.timestamps_enabled? && changed? && !updated_at_changed? && @_touch_record != false`

When...
```
・changed? is false  
=> updated_at will NOT be updated

・updated_at provided by user 
=> updated_at will NOT be updated(override) because changed? is true but !updated_at_changed? is false

・changed? is true and updated_at not provided by user 
 => updated_at will be updated
```

andrykonchin:
Looks pretty good. Thank you for the fix.

PR #526, opened 2022-01-06
Fix a comment of lib/dynamoid/indexes.rb

a5-stable:
fixed a comment of `lib/dynamoid/indexes.rb`.


andrykonchin:
Thank you for the fix!

PR #527, opened 2022-01-20
Use `timestamps_enabled?` in `UpdateFields` and `Upsert`  to prevent `updated_at` from being added to the item

oieioi:
Use `timestamps_enabled?` in `UpdateFields` and `Upsert` to prevent `updated_at` from being added to the item.

Currently, When I use `upsert` or `update_fields`, `updated_at` is added to the item even if I set `table timestamps: false`.


andrykonchin:
Good catch! Thank you for the fix.

PR #533, opened 2022-03-14
Allow custom type for index fields

rahul342:
Since custom types could use allowed index types such as `string`. Allow to be used as keys on indexes.

### DIFF
--- a/lib/dynamoid/indexes.rb
+++ b/lib/dynamoid/indexes.rb
@@ -294,10 +294,10 @@ module Dynamoid
         if @range_key.present?
           range_field_attributes = @dynamoid_class.attributes[@range_key]
           if range_field_attributes.present?
-            range_key_type = range_field_attributes[:type]
-            if Dynamoid::Fields::PERMITTED_KEY_TYPES.include?(range_key_type)
+            range_key_type_dynamodb_type = dynamodb_type(range_field_attributes[:type], range_field_attributes)
+            if Dynamoid::Fields::PERMITTED_KEY_TYPES.include?(range_key_type_dynamodb_type)
               @range_key_schema = {
-                @range_key => PrimaryKeyTypeMapping.dynamodb_type(range_key_type, range_field_attributes)
+                @range_key => range_key_type_dynamodb_type
               }
             else
               errors.add(:range_key, 'Index :range_key is not a valid key type')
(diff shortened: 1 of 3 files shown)


rahul342:
Add #[42171d9](https://github.com/Dynamoid/dynamoid/pull/533/commits/42171d906d4b203a216163981f68b6f7a5e454a5) due to codeclimate build failure on similar duplicate logic.

rahul342:
Add [3212fdb](https://github.com/Dynamoid/dynamoid/pull/533/commits/3212fdb344a15b9582b6cfa6df00f70940b90255) due to codeclimate build failure for duplicate logic and cognitive complexity.

lib/dynamoid/indexes.rb
```diff
@@ -294,10 +294,10 @@ def validate_range_key
         if @range_key.present?
           range_field_attributes = @dynamoid_class.attributes[@range_key]
           if range_field_attributes.present?
-            range_key_type = range_field_attributes[:type]
-            if Dynamoid::Fields::PERMITTED_KEY_TYPES.include?(range_key_type)
+            range_key_type_dynamodb_type = dynamodb_type(range_field_attributes[:type], range_field_attributes)
+            if Dynamoid::Fields::PERMITTED_KEY_TYPES.include?(range_key_type_dynamodb_type)
```
andrykonchin: `PERMITTED_KEY_TYPES ` is a list of Dynamoid types, like date, datetime, and serialized. But `range_key_type_dynamodb_type` is a DynamoDB's type. So they are not completely compatible.

andrykonchin:
Sorry for the delay. I have left a comment.

lib/dynamoid/indexes.rb
rahul342: Thanks for checking! Few options

1. `PERMITTED_KEY_TYPES` doesn't seem to be used anywhere else. We can declare it here and make it a list of dynamodb types instead of dynamoid types
2. Get the underlying type from the custom class similar to `PrimaryKeyTypeMapping.dynamodb_type`. Something like -
```
      def permitted_key_type?(dynamoid_key_type)
        if dynamoid_key_type.is_a?(Class)
          dynamoid_key_type = key_type.respond_to?(:dynamoid_field_type) ? type.dynamoid_field_type : :string
        end
        Dynamoid::Fields::PERMITTED_KEY_TYPES.include?(dynamoid_key_type)
      end

```

Let me know which one is preferrable and also if this feature that we would want to have.

andrykonchin: I would accept both options. But the first one seems a bit simpler. So I would prefer to have a list of allowed DynamoDB types.

rahul342:
Added new commit as discussed in https://github.com/Dynamoid/dynamoid/pull/533#discussion_r829581844

andrykonchin:
Thank you. Looks pretty good.

PR #536, opened 2022-04-01
Gem structure cleanup

pboling:
See Atomic Commits!


.github/workflows/ci.yml
```diff
@@ -7,15 +7,41 @@ on:
...
+          - rails_5_0
+          - rails_5_1
+          - rails_5_2
+          - rails_6_0
+          - rails_6_1
+          - rails_7_0
+        rubygems:
+          - latest
+        bundler:
+          - latest
```
eregon, line 44: Any reason for this? Link: https://github.com/ruby/setup-ruby/issues/441#issuecomment-1374923371

pboling: No specific reason. It seems intuitive to use latest versions, as they will have the most bugfixes related to dependency resolution.

But apparently it is not working that way, and it really has nothing to do with this project.

pboling: The underlying idea is that it normalizes the version as much as possible between builds across many versions of Ruby.  Users of the gem may be keeping pace with rubygems updates in their own builds, and if they aren't nay problems arising from that wouldn't be supported here... It just doesn't make sense to remain stuck on default versions forever.

PR #572, opened 2022-07-25
Proxy kwargs independently from positional args. Closes #571.

### LINKED ISSUE #571: ArgumentError when proxying association method_missing under Ruby 3 due to improper kwarg handling
I have a fix that handles `SingleAssociation`. I looked into fixing `ManyAssociation` too but I'm not sure it's warranted; `Array` doesn't make use of keyword args, so it would be awkward to test.

STR:
```
class Foo
  include Dynamoid::Document
  belongs_to :bar
end

class Bar
  include Dynamoid::Document

  def baz(kwarg1:, kwarg2:)
  end
end

# load a model
foo = ::Foo.find(foo_key, raise_error: false, consistent_read: true)

# follow a belongs_to association to a related model
bar = foo.bar

# call instance method of related model; will be proxied by Dynamoid::Associations::BelongsTo
bar.baz(kwarg1: 'hi')
```

Expected outcome: method is proxied faithfully to `Bar#baz`.

Actual outcome: keyword args are transformed into a hash that is passed as a positional arg
```
ArgumentError: wrong number of arguments (given 1, expected 0; required keywords: kwarg1, kwarg2)
```

xeger:


andrykonchin:
Thank you for the fix.

Could you please exclude all the not relevant formatting changes (e.g. changing '' to "")?

Also I think this change deserves a separate test case, not modifying the existing 

```ruby
it "behaves like the object it is trying to be" do
```

xeger:
Done @andrykonchin; I also added support for proxied-method blocks and `respond_to_missing?` (following best practice). (The blocks were already working, but I added test coverage & made it explicit in the method-missing.)

andrykonchin:
Looks good. Thank you. But there is a failed spec on old Rubies.

xeger:
I fixed the spec. Not a pretty fix, but an efficient one.

xeger:
Closed in favor of #573

PR #573, opened 2022-07-28
Fix ruby3 method proxying rebase. Closes #571 (rebased on master)

### LINKED ISSUE #571: ArgumentError when proxying association method_missing under Ruby 3 due to improper kwarg handling

xeger:
Just like #572, which I've closed due to a branch SNAFU on my fork.

### DIFF
--- a/lib/dynamoid/associations/single_association.rb
+++ b/lib/dynamoid/associations/single_association.rb
@@ -64,15 +64,29 @@ module Dynamoid
         target == other
       end
 
-      # Delegate methods we don't find directly to the target.
-      #
-      # @private
-      # @since 0.2.0
-      def method_missing(method, *args, **kwargs, &block)
-        if target.respond_to?(method)
-          target.send(method, *args, **kwargs, &block)
-        else
-          super
+      if ::RUBY_VERSION < '2.7'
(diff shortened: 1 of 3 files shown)


andrykonchin:
Thank you for the fix! 👍

PR #588, opened 2022-09-06
Support changing type attribute value

bonty:
Hello, Thank you for your maintenance.

I added the feature that supports changing `type` value for STI. It has `sti_name` and `sti_class_for` public methods like `ActiveRecord::Inheritance`.

Additionally, I added `abstract_class` class attribute for restricting not to add `sti_condition` when query or scan request.

Sample code is below.

```ruby
class Animal
  include Dynamoid::Document

  self.abstract_class = true

  field :type

  def self.sti_class_for(type_name)
    case type_name
    when "neko" # neko means cat in Japanese
      Cat
    when "inu" # inu means dog in Japanese
      Dog
    end
  end
end

class Cat < Animal
  def self.sti_name
    "neko"
  end
end

class Dog < Animal
  def self.sti_name
    "inu"
  end
end

Animal.all # This request does not add sti_condition
```

### DIFF
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -787,9 +787,9 @@ module Dynamoid
         condition = {}
         type = @source.inheritance_field
 
-        if @source.attributes.key?(type)
-          class_names = @source.deep_subclasses.map(&:name) << @source.name
-          condition[:"#{type}.in"] = class_names
+        if @source.attributes.key?(type) && !@source.abstract_class?
+          sti_names = @source.deep_subclasses.map(&:sti_name) << @source.sti_name
+          condition[:"#{type}.in"] = sti_names
         end
 
         condition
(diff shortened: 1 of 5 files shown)


andrykonchin:
Hey. Sorry for such a delayed response.

Could you please describe a use case when it's needed custom `type`'s values that shouldn't be derived from class names?

bonty:
@andrykonchin I'm also sorry for the late reply.
I think there are the following use cases for changing `type` values from the class names.

- When records already exist in the table and you want to support STI without data migration.
- When you want to save record capacity for reasons such as long class names

andrykonchin:
Looks good 👍 . Thank you for this improvement.

Initially I was concerned about the naming and some implementation nuances. But lately I realised that it copies the Rails approach and naming.

andrykonchin:
Could you please rebase the branch, it's a bit out of date? I will merge it then.

bonty:
@andrykonchin Thank you for your review! I just rebased the branch.

bonty:
Ah, CI failed. I'll fix this soon.

bonty:
@andrykonchin I've rebased the branch! Please review this again!

andrykonchin:
Thank you!

PR #620, opened 2022-12-20
upsert method is not working. 

jaybee-heon:
TypeCasting#cast_fields and Dumping#dump_field expect to receive attributes(which is hash) not specific value.
I changed it to meet expectation of methods.


andrykonchin:
`Model.attributes` hash contains field declarations - field  type and attributes specified in the `field <name>` method call. `Model.attributes` is updated here https://github.com/Dynamoid/dynamoid/blob/v3.8.0/lib/dynamoid/fields/declare.rb#L16-L18.

If there is anything that you think works in a wrong way - please provide some test cases to reproduce the issue.

PR #621, opened 2022-12-21
Clarify behavior of upsert when document already exists

xeger:
This almost caused an outage in production; thankfully I tested it, but I figured others might appreciate a how-to.


andrykonchin (approved):

andrykonchin:
Thank you!

PR #655, opened 2023-04-21
Migrate to expression attributes for Query and Scan

andrykonchin:
Get rid of legacy `Query`/`Scan`'s attributes.

Use `FilterExpression`, `KeyConditionExpression` and `ProjectionExpression` instead of legacy `ScanFilter`, `QueryFiler` and `AttributesToGet` attributes in `Scan` and `Query` operations.

Changes:
- support multiple conditions in `where` for the same attribute
- removed warnings (introduced in https://github.com/Dynamoid/dynamoid/pull/352)


nbulaj:
Hi @andrykonchin   :wave: 

Just tried to use multiple conditions on the same key:

```ruby
Score.where("timestamp.gte": 1.day.ago).where("timestamp.lte": Time.now).count
```

And also got `Aws::DynamoDB::Errors::ValidationException (KeyConditionExpressions must only contain one condition per key)`

Is see the spec are skipped for it.. But maybe any info on the plans to support multiple conditions?

andrykonchin:
Thank you for reporting the issue.

So DynamoDB allows only one condition for a sort key. I've completely overlooked this restriction. Could you please create a separate issue?

In this particular case `timestamp.between` could be a proper solution to express a `now - 1 day < timestamp < now` condition.

Generally speaking supporting multiple conditions for a sort key isn't as useful as it may seem. DynamoDB supports only subset of operations for searching (in `KeyConditionExpression`, that affects operation cost and performance) - `>`, `>=`, `<`, `<=`, `between`. All the other operators are used in filtering (using `FilterExpression`) that doesn't speed up an operation but reduces HTTP traffic - amount of data that will be send in response. (Actually Dynamoid doesn't distinguish them right now and all the operators are sent as search conditions - so it's a separate bug). So there might be no value in multiple `<` or `>` conditions. It will be just more convenient to build a complex condition. But still it makes sense to allow it.

The only issue with supporting multiple conditions for sort key is that only one of them will be used as a search condition and all the other will be used as filter conditions. This choice will affect performance and cost. For instance there are conditions `a > 0` and `a > 100`. The second one is stricter but if the first one is used for searching and the second one for filtering - operation consumes more units than in the opposite case when the second one is used for searching. And this choice either makes Dynamoid (e.g. takes the first condition for searching and prints warning) or a user explicitly specifying that this particular condition is for searching. The first approach may lead to not optimal performance/cost if user  doesn't notice the warning. The second one makes Dynamoid API more complex (e.g. a new method `#where_keys` could be added) and less generic (because the change is caused of a nuance/limitation of a low level `Query` operation) and doesn't prevent multiple `#where_keys` calls so it  might be unclear which condition to use for searching.

PR #656, opened 2023-04-27
Adds a configuration flag to create table on save

imaximix:
I've added a configuration so you can prevent table creation on save.
We manage tables with something else, so we want to avoid it in production.

I saw that there is another PR that also does this [here](https://github.com/Dynamoid/dynamoid/pull/496)


lib/dynamoid/persistence.rb
```diff
@@ -507,6 +507,11 @@ def persisted?
     # @return [true|false] Whether saving successful or not
     # @since 0.2.0
     def save(options = {})
+      unless Dynamoid.config.create_table_on_save
+        Dynamoid.logger.warn "Table #{table_name} is not created as create_table_enabled is disabled"
+        return false
+      end
+
```
andrykonchin, line 514: Doesn't it prevent model saving at all?

spec/dynamoid/persistence_spec.rb
```diff
@@ -2293,6 +2293,29 @@ def around_update_callback
...
+        Dynamoid.configure do |config|
+          config.create_table_on_save = false
+        end
+      end
+
+      after do
+        Dynamoid.configure do |config|
+          config.create_table_on_save = true
+        end
+      end
```
andrykonchin, line 2308: I wouldn't hardcode option's current value and would just save and restore any value that was set before.

I mean the following:

```ruby
      before do
        Dynamoid.configure do |config|
          @create_table_on_save_old = config.create_table_on_save
          config.create_table_on_save = false
        end
      end

      after do
        Dynamoid.configure do |config|
          config.create_table_on_save = @create_table_on_save_old
        end
      end
```

spec/dynamoid/persistence_spec.rb
```diff
@@ -2293,6 +2293,29 @@ def around_update_callback
...
+
+      it 'does not create table' do
+        model = klass.new
+
+        expect(klass).not_to receive(:create_table)
+
+        model.save
+      end
+    end
+
```
andrykonchin, line 2318: I suppose we need this test for each method that persists a model, e.g. `#update_attributes` etc.

lib/dynamoid/persistence.rb
imaximix: Yes, you are right. These unless are hard on the brain.

lib/dynamoid/persistence.rb
```diff
@@ -507,14 +507,22 @@ def persisted?
...
       create_or_update = new_record? ? :create : :update
 
       run_callbacks(:save) do
         run_callbacks(create_or_update) do
           Save.call(self, touch: options[:touch])
         end
       end
+    rescue Aws::DynamoDB::Errors::ResourceNotFoundException => error
+      Dynamoid.logger.error(error)
+      return false
```
andrykonchin, line 525: I think it's better to have this exception visible to user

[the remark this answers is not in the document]
andrykonchin: I think it's better to have this exception visible to user

spec/dynamoid/persistence_spec.rb
imaximix: Also we are not sure that testing `#update_attributes` and co provides that much value because they both call save. What do you think?

andrykonchin: It's OK to leave it as is.

I will add specs on my own to be sure behaviour isn't changed in case of some future refactoring.

andrykonchin (approved):

andrykonchin:
Could you please squash commits into a single one after CI passes successfully?

lib/dynamoid/persistence.rb
```diff
@@ -507,7 +507,12 @@ def persisted?
     # @return [true|false] Whether saving successful or not
     # @since 0.2.0
     def save(options = {})
-      self.class.create_table(sync: true)
+      if Dynamoid.config.create_table_on_save
+        self.class.create_table(sync: true)
+      else
+        Dynamoid.logger.warn "Table #{self.class.table_name} is not created as create_table_enabled is disabled"
+      end
+
```
imaximix, line 515: @andrykonchin What do you think about removing the warn? it's gonna show up on every save. And we have the exception anyway

andrykonchin: Yeah, agree.

TBH I would consider implicit table creation as an auxiliary feature, that is used in development/test environments and will be disabled on production. So if somebody expects a warning about unusual behaviour - I suppose it should be a warning when a table is created implicitly. But I believe we should rely on user experience and add warning when users ask for it.

andrykonchin (approved):

lib/dynamoid/persistence.rb
imaximix: great, i'll remove the log then

andrykonchin:
Thank you!

Issue #277, opened 2018-07-19
When I call Dynamoid.included_models, the returned list seems to be missing some models


rhuang:
Any idea why this might be happening? All my `Dynamoid` models are in various nested folders under `/app/models/*`. I'm already using all these `Dynamoid` classes in production - I'm just trying to setup proper test cleanup and I seem to be missing tables during creation.

andrykonchin:
Dynamoid adds model class  to the `Dynamoid.included_models` list only when a file with model declaration is required with `require` or `load`.

Looks like in test environment some files aren't required.

How do you reproduce the issue?

rhuang:
In `spec_helpers.rb` (my `RSpec` config file), I call `Dynamoid.included_models` in `config.before(:suite)` to create tables, in preparation for tests that actually put things into Dynamo. I noticed that `included_models` is missing a bunch of models, and I can't seem to see a pattern to it. I also tried explicitly doing `require` in `spec_helpers.rb` for the missing model files, but it doesn't seem to make a difference. This is also reproducible if I execute `Dynamoid.included_models` in the `Rails` console.

andrykonchin:
> I guess I am just curious why it wasn't able to find them in the first place.

It would be helpful if you were able to provide minimized example to reproduce the issue. Now I can you guess.

Document occurs in `Dynamoid.included_models` only after loading the file with definition of the class (look at [this](https://github.com/Dynamoid/Dynamoid/blob/12b874c0d721f5993b0a2f6839dce53efe3c0973/lib/dynamoid/document.rb#L16)).
If class is a subclass of another document (some sort of STI) it doesn't occur on the `Dynamoid.included_models` but it's OK as far as both classes shares the same DynamoDB table.

pboling:
The test environment does not eager load all models, unlike production.  So `included_models` will grow as the tests run.  It can't be used for bootstrapping of the environment unless you manually load all the models with `require` or `load`.  There is nothing this project can do about this.  The test env is simply a different beast than production.

misham:
I'm having a similar issue, using the test bootstrapping code that re-creates all the tables for every test run.  The problem is that `Dynamoid.included_models` is empty every time, until that particular model is being tested.  That means that `DynamoidReset` module is not creating any tables, after it deletes them.

The workaround I have is including the models into `DynamoidReset` module.  Ideally, there would be a way to eager load Dynamoid models instead.

Any suggestions on how to accomplish this?

[...]

PR #687, opened 2023-10-10
Update README.md

dunkOnIT:
This PR adds a note about requiring model files when running RSpec tests, based on #277. 

I believe this should be documented, as it is not obvious how included_models is set, or that it would be empty.


andrykonchin:
Thank you!

PR #688, opened 2023-10-12
transactions multi-table MVP

ckhsponge:
This is a first attempt at dynamodb transactions that can span tables. Create, update, upsert and delete are supported. See README_transact.md for examples of the proposed API.

Tests have not yet been implemented.

### DIFF
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
@@ -8,6 +8,7 @@ require_relative 'aws_sdk_v3/batch_get_item'
 require_relative 'aws_sdk_v3/item_updater'
 require_relative 'aws_sdk_v3/table'
 require_relative 'aws_sdk_v3/until_past_table_status'
+require_relative 'aws_sdk_v3/transact'
 
 module Dynamoid
   # @private
@@ -287,6 +288,10 @@ module Dynamoid
         end
       rescue Aws::DynamoDB::Errors::ConditionalCheckFailedException => e
         raise Dynamoid::Errors::ConditionalCheckFailedException, e
+      end
+
+      def transact_write_items(items)
+        Transact.new(client).transact_write_items(items)
       end
 
       # Create a table on DynamoDB. This usually takes a long time to complete.
--- a/lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
+++ b/lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# Prepare all the actions of the transaction for sending to the AWS SDK.
+module Dynamoid
+  module AdapterPlugin
+    class AwsSdkV3
+      class Transact
+        attr_reader :client
+
+        def initialize(client)
+          @client = client
+        end
+
+        # Perform all of the item actions in a single transaction.
+        #
+        # @param [Array] items of type Dynamoid::Transaction::Action or
+        # any other object whose to_h is a transact_item hash
+        #
+        def transact_write_items(items)
+          transact_items = items.map(&:to_h)
+          params = {
+            transact_items: transact_items,
+            return_consumed_capacity: 'TOTAL',
+            return_item_collection_metrics: 'SIZE'
+          }
+          client.transact_write_items(params) # returns this
+        end
+      end
+    end
+  end
+end
--- a/spec/dynamoid/transaction_write/context.rb
+++ b/spec/dynamoid/transaction_write/context.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+# Dynamoid.config.logger.level = :debug
+
+RSpec.shared_context 'transaction_write' do
+  let(:klass) do
+    new_class(class_name: 'Document') do
+      field :name
+    end
+  end
+
+  let(:klass_with_composite_key) do
+    new_class(class_name: 'Cat') do
+      range :age, :integer
+      field :name
+    end
+  end
+
+  let(:klass_with_callbacks) do
+    new_class(class_name: 'Dog') do
+      field :name
+
+      before_save { print 'saving ' }
+      after_save { print 'saved ' }
+      before_create { print 'creating ' }
+      after_create { print 'created ' }
+      before_update { print 'updating ' }
+      after_update { print 'updated ' }
+      before_destroy { print 'destroying ' }
+      after_destroy { print 'destroyed ' }
+      before_validation { print 'validating ' }
+      after_validation { print 'validated ' }
+    end
+  end
+
+  let(:klass_with_around_callbacks) do
+    new_class(class_name: 'Mouse') do
+      field :name
+
+      around_save :around_save_callback
+      around_create :around_create_callback
+      around_update :around_update_callback
+      around_destroy :around_destroy_callback
+      # no around_validation callback exists
+
+      def around_save_callback
+        print 'saving '
+        yield
+        print 'saved '
+      end
+
+      def around_create_callback
+        print 'creating '
+        yield
+        print 'created '
+      end
+
+      def around_update_callback
+        print 'updating '
+        yield
+        print 'updated '
+      end
+
+      def around_destroy_callback
+        print 'destroying '
+        yield
+        print 'destroyed '
+      end
+    end
+  end
+
+  let(:klass_with_validation) do
+    new_class do
+      field :name
+      validates :name, length: { minimum: 4 }
+    end
+  end
+end
--- a/spec/dynamoid/transaction_write/create_spec.rb
+++ b/spec/dynamoid/transaction_write/create_spec.rb
@@ -1,0 +1,200 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require_relative 'context'
+
+# Dynamoid.config.logger.level = :debug
+
+describe Dynamoid::TransactionWrite, '.create' do
+  include_context 'transaction_write'
+
+  context 'creates' do
+    context 'simple primary key' do
+      before do
+        klass.create_table
+      end
+
+      it 'with attribute in constructor' do
+        obj1 = klass.new(name: 'one')
+        expect(obj1.persisted?).to eql(false)
+        described_class.execute do |txn|
+          txn.create! obj1
+        end
+        expect(obj1.persisted?).to eql(true)
+        obj1_found = klass.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('one')
+      end
+
+      it 'with attribute in transaction' do
+        obj2 = klass.new
+        described_class.execute do |txn|
+          txn.create! obj2, name: 'two'
+        end
+        obj2_found = klass.find(obj2.id)
+        expect(obj2_found).to eql(obj2)
+        expect(obj2_found.name).to eql('two')
+      end
+
+      it 'with class constructed in transaction' do
+        obj3 = nil
+        described_class.execute do |txn|
+          obj3 = txn.create! klass, name: 'three'
+        end
+        obj3_found = klass.find(obj3.id)
+        expect(obj3_found).to eql(obj3)
+        expect(obj3_found.name).to eql('three')
+      end
+    end
+
+    context 'composite key' do
+      before do
+        klass_with_composite_key.create_table
+      end
+
+      it 'with attribute in constructor' do
+        obj1 = klass_with_composite_key.new(name: 'one', age: 1)
+        described_class.execute do |txn|
+          txn.create! obj1
+        end
+        obj1_found = klass_with_composite_key.find(obj1.id, range_key: 1)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('one')
+      end
+
+      it 'with attribute in transaction' do
+        obj2 = klass_with_composite_key.new
+        described_class.execute do |txn|
+          txn.create! obj2, name: 'two', age: 2
+        end
+        obj2_found = klass_with_composite_key.find(obj2.id, range_key: 2)
+        expect(obj2_found).to eql(obj2)
+        expect(obj2_found.name).to eql('two')
+      end
+
+      it 'with class constructed in transaction' do
+        obj3 = nil
+        described_class.execute do |txn|
+          obj3 = txn.create! klass_with_composite_key, name: 'three', age: 3
+        end
+        obj3_found = klass_with_composite_key.find(obj3.id, range_key: 3)
+        expect(obj3_found).to eql(obj3)
+        expect(obj3_found.name).to eql('three')
+      end
+    end
+
+    it 'creates timestamps' do
+      klass.create_table
+      obj1 = klass.new(name: 'one')
+      expect(obj1.created_at).to be_nil
+      expect(obj1.updated_at).to be_nil
+      described_class.execute do |txn|
+        txn.create! obj1
+      end
+      obj1_found = klass.find(obj1.id)
+      expect(obj1_found.created_at.to_f).to be_within(1.seconds).of Time.now.to_f
+      expect(obj1_found.updated_at.to_f).to be_within(1.seconds).of Time.now.to_f
+    end
+
+    context 'validates' do
+      before do
+        klass_with_validation.create_table
+      end
+
+      it 'does not create when invalid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        described_class.execute do |txn|
+          expect(txn.create(obj1)).to eql(false)
+        end
+        expect(obj1.id).to be_nil
+      end
+
+      it 'rolls back when invalid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        obj2 = klass_with_validation.new(name: 'twotwo')
+        described_class.execute do |txn|
+          expect(txn.create(obj1)).to eql(false)
+          expect(txn.create(obj2)).to be_present
+        end
+        expect(obj1.id).to be_nil
+        expect(klass_with_validation).to exist(obj2.id)
+      end
+
+      it 'succeeds when valid' do
+        obj1 = klass_with_validation.new(name: 'oneone')
+        described_class.execute do |txn|
+          expect(txn.create(obj1)).to be_present
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+      end
+
+      it 'raises DocumentNotValid when not valid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        expect {
+          described_class.execute do |txn|
+            txn.create! obj1
+          end
+        }.to raise_error(Dynamoid::Errors::DocumentNotValid)
+        expect(obj1.id).to be_nil # hash key should NOT be auto-generated if validation fails
+      end
+
+      it 'rolls back and raises DocumentNotValid when not valid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        obj2 = klass_with_validation.new(name: 'twotwo')
+        expect {
+          described_class.execute do |txn|
+            txn.create! obj2
+            txn.create! obj1
+          end
+        }.to raise_error(Dynamoid::Errors::DocumentNotValid)
+        expect(obj1.id).to be_nil
+        expect(obj2.id).to be_present
+        expect(klass_with_validation).not_to exist(obj2.id)
+      end
+
+      it 'does not raise exception when valid' do
+        obj1 = klass_with_validation.new(name: 'oneone')
+        described_class.execute do |txn|
+          txn.create!(obj1)
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+      end
+
+      it 'does not raise exception when skipping validation' do
+        obj1 = klass_with_validation.new(name: 'one')
+        described_class.execute do |txn|
+          # this use is infrequent, normal entry is from save!(obj, options)
+          txn.create!(obj1, {}, skip_validation: true)
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('one')
+      end
+    end
+
+    it 'uses callbacks' do
+      klass_with_callbacks.create_table
+      expect {
+        described_class.execute do |txn|
+          txn.create! klass_with_callbacks.new(name: 'two')
+        end
+      }.to output('validating validated saving creating created saved ').to_stdout
+    end
+
+    it 'uses around callbacks' do
+      klass_with_around_callbacks.create_table
+      expect {
+        described_class.execute do |txn|
+          txn.create! klass_with_around_callbacks.new(name: 'two')
+        end
+      }.to output('saving creating created saved ').to_stdout
+    end
+  end
+end
--- a/spec/dynamoid/transaction_write/put_spec.rb
+++ b/spec/dynamoid/transaction_write/put_spec.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require_relative 'context'
+
+# Dynamoid.config.logger.level = :debug
+
+describe Dynamoid::TransactionWrite, '.put' do # 'put' is actually an update
+  include_context 'transaction_write'
+
+  # a 'put' is a create that overwrites existing records if present
+  context 'puts' do
+    context 'simple primary key' do
+      before do
+        klass.create_table
+      end
+
+      it 'fails without skip_existence_check' do
+        obj1 = klass.create!(name: 'one')
+        expect {
+          described_class.execute do |txn|
+            txn.create! klass, id: obj1.id, name: 'oneo'
+          end
+        }.to raise_exception(Aws::DynamoDB::Errors::TransactionCanceledException)
+      end
+
+      it 'can overwrite with attributes' do
+        obj1 = klass.create!(name: 'one')
+        expect(klass.find(obj1.id)).to eql(obj1) # it was created
+        described_class.execute do |txn|
+          txn.create! klass, { id: obj1.id, name: 'oneone' }, { skip_existence_check: true }
+        end
+        obj1_found = klass.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone') # overwritten
+      end
+
+      it 'can overwrite with instance' do
+        obj1 = klass.create!(name: 'one')
+        expect(klass.find(obj1.id)).to eql(obj1) # it was created
+        described_class.execute do |txn|
+          # this use is infrequent, normal entry is from save!(obj, options)
+          txn.create! klass.new(id: obj1.id, name: 'oneone'), {}, { skip_existence_check: true }
+        end
+        obj1_found = klass.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone') # overwritten
+      end
+    end
+
+    context 'composite key' do
+      before do
+        klass_with_composite_key.create_table
+      end
+
+      it 'fails without skip_existence_check' do
+        obj1 = klass_with_composite_key.create!(name: 'one', age: 1)
+        expect {
+          described_class.execute do |txn|
+            txn.create! klass_with_composite_key, id: obj1.id, age: 1, name: 'oneo'
+          end
+        }.to raise_exception(Aws::DynamoDB::Errors::TransactionCanceledException)
+      end
+
+      it 'can overwrite with attributes' do
+        obj1 = klass_with_composite_key.create!(name: 'one', age: 1)
+        obj2 = nil
+        described_class.execute do |txn|
+          txn.create! klass_with_composite_key, { id: obj1.id, age: 1, name: 'oneone' },
+                      { skip_existence_check: true }
+          obj2 = txn.create! klass_with_composite_key, id: obj1.id, age: 2, name: 'two'
+        end
+        obj1_found = klass_with_composite_key.find(obj1.id, range_key: 1)
+        obj2_found = klass_with_composite_key.find(obj1.id, range_key: 2)
+        expect(obj1_found).to eql(obj1)
+        expect(obj2_found).to eql(obj2)
+        expect(obj1_found.name).to eql('oneone')
+        expect(obj2_found.name).to eql('two')
+      end
+
+      it 'can overwrite with instance' do
+        obj1 = klass_with_composite_key.create!(name: 'one', age: 1)
+        obj2 = nil
+        described_class.execute do |txn|
+          # this use is infrequent, normal entry is from save!(obj, options)
+          txn.create! klass_with_composite_key.new(id: obj1.id, age: 1, name: 'oneone'), {},
+                      { skip_existence_check: true }
+          obj2 = txn.create! klass_with_composite_key.new(id: obj1.id, age: 2, name: 'two')
+        end
+        obj1_found = klass_with_composite_key.find(obj1.id, range_key: 1)
+        obj2_found = klass_with_composite_key.find(obj1.id, range_key: 2)
+        expect(obj1_found).to eql(obj1)
+        expect(obj2_found).to eql(obj2)
+        expect(obj1_found.name).to eql('oneone')
+        expect(obj2_found.name).to eql('two')
+      end
+    end
+  end
+end
--- a/spec/dynamoid/transaction_write/save_spec.rb
+++ b/spec/dynamoid/transaction_write/save_spec.rb
@@ -1,0 +1,190 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require_relative 'context'
+
+# Dynamoid.config.logger.level = :debug
+
+describe Dynamoid::TransactionWrite, '.save' do # 'save' is an update or create
+  include_context 'transaction_write'
+
+  # a 'save' does an update or create depending on if the record is new or not
+  context 'saves' do
+    context 'simple primary key' do
+      before do
+        klass.create_table
+      end
+
+      it 'with an update' do
+        obj1 = klass.create!(name: 'one')
+        described_class.execute do |txn|
+          obj1.name = 'oneone'
+          txn.save! obj1
+        end
+        obj1_found = klass.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+      end
+
+      it 'with a create' do
+        obj2 = klass.new(name: 'two')
+        described_class.execute do |txn|
+          txn.save! obj2
+        end
+        obj2_found = klass.find(obj2.id)
+        expect(obj2_found).to eql(obj2)
+        expect(obj2_found.name).to eql('two')
+      end
+    end
+
+    context 'composite key' do
+      before do
+        klass_with_composite_key.create_table
+      end
+
+      it 'with an update' do
+        obj1 = klass_with_composite_key.create!(name: 'one', age: 1)
+        described_class.execute do |txn|
+          obj1.name = 'oneone'
+          txn.save! obj1
+        end
+        obj1_found = klass_with_composite_key.find(obj1.id, range_key: 1)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+      end
+
+      it 'with a create' do
+        obj2 = klass_with_composite_key.new(name: 'two', age: 2)
+        described_class.execute do |txn|
+          txn.save! obj2
+        end
+        obj2_found = klass_with_composite_key.find(obj2.id, range_key: 2)
+        expect(obj2_found).to eql(obj2)
+        expect(obj2_found.name).to eql('two')
+      end
+    end
+
+    context 'validates' do
+      before do
+        klass_with_validation.create_table
+      end
+
+      it 'does not save when invalid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        described_class.execute do |txn|
+          expect(txn.save(obj1)).to eql(false)
+        end
+        expect(obj1.id).to be_nil
+      end
+
+      it 'allows partial save when a record in the transaction is invalid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        obj2 = klass_with_validation.create!(name: 'twolong')
+        obj2.name = 'twotwo'
+        described_class.execute do |txn|
+          expect(txn.save(obj2)).to be_truthy
+          expect(txn.save(obj1)).to eql(false)
+        end
+        expect(obj1.id).to be_nil
+        obj2_found = klass_with_validation.find(obj2.id)
+        expect(obj2_found.name).to eql('twotwo')
+      end
+
+      it 'saves when valid' do
+        obj1 = klass_with_validation.new(name: 'oneone')
+        described_class.execute do |txn|
+          expect(txn.save(obj1)).to be_present
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+      end
+
+      it 'raises DocumentNotValid when not valid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        expect {
+          described_class.execute do |txn|
+            txn.save! obj1
+          end
+        }.to raise_error(Dynamoid::Errors::DocumentNotValid)
+        expect(obj1.id).to be_nil
+      end
+
+      it 'rolls back and raises DocumentNotValid when not valid' do
+        obj1 = klass_with_validation.new(name: 'one')
+        obj2 = klass_with_validation.create!(name: 'twolong')
+        obj2.name = 'twotwo'
+        expect {
+          described_class.execute do |txn|
+            txn.save! obj2
+            txn.save! obj1
+          end
+        }.to raise_error(Dynamoid::Errors::DocumentNotValid)
+        expect(obj1.id).to be_nil
+        obj2_found = klass_with_validation.find(obj2.id)
+        expect(obj2_found.name).to eql('twolong')
+      end
+
+      it 'does not raise exception when valid' do
+        obj1 = klass_with_validation.new(name: 'oneone')
+        described_class.execute do |txn|
+          txn.save!(obj1)
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+      end
+
+      it 'does not raise exception when skipping validation' do
+        obj1 = klass_with_validation.new(name: 'one')
+        described_class.execute do |txn|
+          txn.save!(obj1, skip_validation: true)
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('one')
+      end
+    end
+
+    context 'callabacks' do
+      it 'uses callbacks' do
+        klass_with_callbacks.create_table
+        expect {
+          described_class.execute do |txn|
+            txn.save! klass_with_callbacks.new(name: 'two')
+          end
+        }.to output('validating validated saving creating created saved ').to_stdout
+      end
+
+      it 'uses around callbacks' do
+        klass_with_around_callbacks.create_table
+        expect {
+          described_class.execute do |txn|
+            txn.save! klass_with_around_callbacks.new(name: 'two')
+          end
+        }.to output('saving creating created saved ').to_stdout
+      end
+
+      it 'cak skip callbacks' do
+        klass_with_callbacks.create_table
+        expect {
+          described_class.execute do |txn|
+            txn.save! klass_with_callbacks.new(name: 'two'), skip_callbacks: true
+          end
+        }.to output('validating validated ').to_stdout # ActiveModel runs validation callbacks when we validate
+      end
+
+      it 'cak skip callbacks and validation' do
+        klass_with_callbacks.create_table
+        expect {
+          described_class.execute do |txn|
+            txn.save! klass_with_callbacks.new(name: 'two'), skip_callbacks: true, skip_validation: true
+          end
+        }.to output('').to_stdout
+      end
+    end
+  end
+end
--- a/spec/dynamoid/transaction_write/update_spec.rb
+++ b/spec/dynamoid/transaction_write/update_spec.rb
@@ -1,0 +1,229 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require_relative 'context'
+
+# Dynamoid.config.logger.level = :debug
+
+describe Dynamoid::TransactionWrite, '.update' do
+  include_context 'transaction_write'
+
+  context 'updates' do
+    context 'simple primary key' do
+      before do
+        klass.create_table
+      end
+
+      it 'with attribute outside transaction' do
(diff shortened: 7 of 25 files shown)


andrykonchin (commented):
I've left some comments. It isn't a call to action to make any change but is a way to get you more familiar with the project structure and its conventions.

lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
```diff
@@ -0,0 +1,158 @@
...
+        # Format the create action as a put. The document must NOT already exist.
+        def transact_item_create(item)
+          model_class = item[:model_class]
+          raise "missing model_class" unless model_class
+          attributes = item[:attributes].clone
+          transact_add_timestamps(model_class, attributes) unless item[:skip_timestamps]
+          attributes[model_class.hash_key] ||= SecureRandom.uuid
+          attributes = Dynamoid::Dumping.dump_attributes(attributes, model_class.attributes)
+          condition = "attribute_not_exists(#{model_class.hash_key})"
+          condition += " and attribute_not_exists(#{model_class.range_key})" if model_class.range_key? # needed?
```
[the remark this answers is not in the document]
ckhsponge: Indeed. Thanks for mentioning.

lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
```diff
@@ -0,0 +1,158 @@
...
+        # keys in item[:attributes] are NOT sanitized for injection attacks
+        def transact_item_update(item)
+          model_class = item[:model_class]
+          raise "missing model_class" unless model_class
+          attributes = item[:attributes].clone
+
+          # ensure types are correct for incoming attributes
+          all_types_string = attributes.keys.map{|k| [k,{type: :string}]}.to_h
+          transact_add_timestamps(model_class, attributes, skip_created_at: true) unless item[:skip_timestamps]
+          attributes.delete(model_class.inheritance_field) # do not update type field
```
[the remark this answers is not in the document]
ckhsponge: Are you wondering if ActiveRecord timestamp columns can be renamed? I don't believe they can be but that doesn't mean dynamoid couldn't :)
https://api.rubyonrails.org/classes/ActiveRecord/Timestamp.html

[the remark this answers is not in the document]
andrykonchin: I mean that in Rails it might be allowed to change values of "system" columns explicitly (timestamps, inheritance column, lock version column), e.g. with `user.update_attributes(type: 'Employee')`. But in the code above the `type` attribute is filtered out.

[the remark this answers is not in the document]
ckhsponge: Got it. I think I was seeing an error from dynamoid when trying to set the type but I also think we can remove that line so it shouldn't be an issue. Thanks.

ckhsponge:
@andrykonchin I made some updates to the transaction API and now there is no longer a mess of hashes being passed around. Your feedback would be appreciated. Some minor cleanup and tests are still needed.

lib/dynamoid/transaction/create.rb
```diff
@@ -0,0 +1,37 @@
...
+      def run_callbacks
+        (yield; return) unless model
+        model.run_callbacks(:save) do
+          model.run_callbacks(:create) do
+            model.run_callbacks(:validate) do
+              yield
+            end
+          end
+        end
+      end
```
andrykonchin, line 16: As a user I would expect that `before` callback are called before persisting and `after` callback after it. So in `after` callbacks model persistence is guarantied.

lib/dynamoid/transaction/action.rb
```diff
@@ -0,0 +1,58 @@
+
+module Dynamoid
+  module Transaction
+    class Action
+      attr_accessor :model, :model_class, :attributes, :options
+
+      def initialize(model_or_model_class, attributes = {}, options = {})
+        if model_or_model_class.respond_to?('new_record?')
```
andrykonchin, line 8:
```suggestion
        if model_or_model_class.is_a? Dynamoid::Document
```

lib/dynamoid/transaction/create.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'action'
+
+module Dynamoid
+  module Transaction
+    class Create < Action
+
+      def run_callbacks
+        (yield; return) unless model
```
andrykonchin, line 8: I assume it's a branch for `create(User, attributes)`. Shouldn't we run callbacks as well in this case?

lib/dynamoid/transaction/create.rb
```diff
@@ -0,0 +1,37 @@
...
+          end
+        end
+      end
+
+      def to_h
+        item = attributes.clone
+        item = add_timestamps(item)
+        item = model.attributes.merge(item) if model
+        item[model_class.hash_key] ||= SecureRandom.uuid
+        item = Dynamoid::Dumping.dump_attributes(item, model_class.attributes)
```
andrykonchin, line 23: This code assumes Symbol keys in `attributes`. Dynamoid accepts String keys as well usually.

andrykonchin: Field values should also be type casted, not only dumped so use could write `create(age: "12")`.

lib/dynamoid/transaction/update.rb
```diff
@@ -0,0 +1,68 @@
...
+      def to_h
+        item = attributes.clone
+
+        # ensure types are correct for incoming attributes
+        all_types_string = item.keys.map{|k| [k,{type: :string}]}.to_h
+        item = add_timestamps(item, skip_created_at: true)
+        item = model.attributes.merge(item) if model
+        # TODO only update dirty attributes, deleting the inheritance field will not be needed
+        item.delete(model_class.inheritance_field) # do not update type field
+        item = Dynamoid::Dumping.dump_attributes(item, all_types_string.merge(model_class.attributes))
```
andrykonchin, line 27: What is the purpose of `all_types_string`? To accept not declared fields? It isn't supported in Dynamoid intentionally, `UnknownAttribute` is raised in this case.

lib/dynamoid/transaction/update.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def to_h
+        item = attributes.clone
+
+        # ensure types are correct for incoming attributes
+        all_types_string = item.keys.map{|k| [k,{type: :string}]}.to_h
+        item = add_timestamps(item, skip_created_at: true)
+        item = model.attributes.merge(item) if model
+        # TODO only update dirty attributes, deleting the inheritance field will not be needed
+        item.delete(model_class.inheritance_field) # do not update type field
```
andrykonchin, line 26: User is allowed to update any system attribute like `created_at` or `type`. So `inheritance_field` shouldn't be deleted.

lib/dynamoid/transaction/update.rb
```diff
@@ -0,0 +1,68 @@
...
+        item = Dynamoid::Dumping.dump_attributes(item, all_types_string.merge(model_class.attributes))
+
+        # set 'key' that is used to look up record for updating
+        hash_key = item.delete(model_class.hash_key)
+        key = {model_class.hash_key => hash_key}
+        key[model_class.range_key] = item.delete(model_class.range_key) if model_class.range_key?
+
+        # e.g. "SET #updated_at = :updated_at ADD record_count :i"
+        update_expression = options[:update_expression] || ("SET " + item.keys.map{ |k| "##{k} = :#{k}" }.join(", "))
```
andrykonchin, line 35: **suggestion:** it makes sense to use short generated placeholders like `a0`, `b1` etc to have expression string shorter. 

https://github.com/Dynamoid/dynamoid/blob/6d53f100f0478bc04c7c45646771182df603fdc8/lib/dynamoid/adapter_plugin/aws_sdk_v3/query.rb#L58-L59

lib/dynamoid/transaction/update.rb
```diff
@@ -0,0 +1,68 @@
...
+          update: {
+            key: key,
+            table_name: model_class.table_name,
+            update_expression: update_expression,
+            expression_attribute_values: expression_attribute_values
+          }
+        }
+        result[:update][:expression_attribute_names] = expression_attribute_names if expression_attribute_names.present?
+        result[:update][:condition_expression] = condition_expression unless item[:skip_existence_check]
+        puts "UPDATE #{result}"
```
andrykonchin, line 63: For debugging all requests/responses to DynamoDb could be logged with a special formatter and setting log level to `debug` like it's done in tests

https://github.com/Dynamoid/dynamoid/blob/6d53f100f0478bc04c7c45646771182df603fdc8/spec/spec_helper.rb#L76

https://github.com/Dynamoid/dynamoid/blob/6d53f100f0478bc04c7c45646771182df603fdc8/spec/support/log_level.rb#L3-L11

lib/dynamoid/transaction/upsert.rb
```diff
@@ -0,0 +1,33 @@
...
+        item = model.attributes.merge(item) if model
+        item[model_class.hash_key] ||= SecureRandom.uuid
+        item = Dynamoid::Dumping.dump_attributes(item, model_class.attributes)
+        {
+          put: {
+            item: sanitize_item(item),
+            table_name: model_class.table_name,
+          }
+        }
+      end
```
andrykonchin, line 30: Class-level methods like `create`/`update`/`upsert` accepts keys and attributes as separate parameters:

```ruby
 User.upsert('1', age: 26)
```

This way we can find the proper item in DynamoDB even when user updates sort key for instance (can he change a partition key? probably he can)

andrykonchin (commented):
Looks pretty good!

lib/dynamoid/transaction/create.rb
ckhsponge: I believe since these are nested all of the before callbacks will be run first, then persisted, then after callbacks are run. Am I incorrect?

lib/dynamoid/transaction/executor.rb
```diff
@@ -0,0 +1,78 @@
...
+        actions = actions.clone
+        action = actions.shift # remove first item
+        if action.respond_to?(:run_callbacks)
+          action.run_callbacks do
+            call_with_callbacks(actions) # recursive call with remaining items
+          end
+        else
+          call_with_callbacks(actions)
+        end
+      end
```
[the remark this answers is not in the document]
ckhsponge: Recursion is used so `before` callbacks are run on every model, then models are persisted, then all `after` callbacks are run.
```
Transaction do
  save(user)
  save(tag)
end
```
unwinds to:
before `user` callbacks
before `tag` callbacks
persist
after `tag` callbacks
after `user` callbacks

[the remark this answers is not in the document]
andrykonchin: Ah, I see. Right 👍

lib/dynamoid/transaction/create.rb
andrykonchin: 👍

lib/dynamoid/transaction/executor.rb
```diff
@@ -0,0 +1,78 @@
...
+        @skip_callbacks = !!options[:skip_callbacks]
+        @skip_validation = !!options[:skip_validation]
+        @actions = actions
+        instance_eval &block if block
+      end
+
+      def call
+        validate! unless @skip_validation
+        call_with_callbacks(@actions)
+      end
```
andrykonchin, line 21: It seems validation could be done twice? Here and in actions?

lib/dynamoid/transaction/create.rb
```diff
@@ -0,0 +1,37 @@
...
+        condition = "attribute_not_exists(#{model_class.hash_key})"
+        condition += " and attribute_not_exists(#{model_class.range_key})" if model_class.range_key? # needed?
+        {
+          put: {
+            item: sanitize_item(item),
+            table_name: model_class.table_name,
+            condition_expression: condition,
+          }
+        }
+      end
```
andrykonchin, line 34: After persisting a new model should become `persisted?` and modified fields should become not-changed:

https://github.com/Dynamoid/dynamoid/blob/6d53f100f0478bc04c7c45646771182df603fdc8/lib/dynamoid/persistence/save.rb#L46
https://github.com/Dynamoid/dynamoid/blob/6d53f100f0478bc04c7c45646771182df603fdc8/lib/dynamoid/dirty.rb#L47

lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
```diff
@@ -0,0 +1,158 @@
...
+            return_item_collection_metrics: 'SIZE'
+          }
+          return client.transact_write_items(params)
+        end
+
+        private
+        # Format the create action as a put. The document must NOT already exist.
+        def transact_item_create(item)
+          model_class = item[:model_class]
+          raise "missing model_class" unless model_class
```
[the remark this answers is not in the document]
ckhsponge: These were removed.

lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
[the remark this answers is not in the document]
ckhsponge: Removed this.

lib/dynamoid/adapter_plugin/aws_sdk_v3/transact.rb
```diff
@@ -0,0 +1,29 @@
...
+        #
+        def transact_write_items(items)
+          transact_items = items.map(&:to_h)
+          params = {
+            transact_items: transact_items,
+            return_consumed_capacity: 'TOTAL',
+            return_item_collection_metrics: 'SIZE'
+          }
+          puts "transact_write_items #{params}"
+          return client.transact_write_items(params)
```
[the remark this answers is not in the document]
ckhsponge: Removed the 'return'. I believe the AWS SDK checks the format so I'm not sure we need to.

lib/dynamoid/transaction/executor.rb
```diff
@@ -0,0 +1,78 @@
...
+        @actions << Dynamoid::Transaction::Upsert.new(model_or_model_class, attributes, options)
+      end
+
+      def update(model_or_model_class, attributes = {}, options = {})
+        @actions << Dynamoid::Transaction::Update.new(model_or_model_class, attributes, options)
+      end
+
+      def delete(model_or_model_class, attributes = {}, options = {})
+        @actions << Dynamoid::Transaction::Delete.new(model_or_model_class, attributes, options)
+      end
```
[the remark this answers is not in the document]
ckhsponge: I was able to make the instance and class invocations very similar since the class invocation will always create or find an instance first. However, none of these accept blocks at the moment so that could change.

lib/dynamoid/transaction/executor.rb
```diff
@@ -0,0 +1,78 @@
...
+        validate! unless @skip_validation
+        call_with_callbacks(@actions)
+      end
+
+      # throws Dynamoid::Errors::DocumentNotValid if model is present and not valid
+      # use skip_validation: true to skip
+      def validate!
+        @actions.each do |action|
+          next if !action.is_a?(Action) || !(model = action.model) || action.skip_validation?
+          raise Dynamoid::Errors::DocumentNotValid.new(model) unless model.valid?
```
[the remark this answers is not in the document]
ckhsponge: Yes, non-raising is now supported e.g. update() vs. update!().

lib/dynamoid/transaction/create.rb
ckhsponge: Yes, this is done now.

lib/dynamoid/transaction/action.rb
ckhsponge: Yes, this is better.

lib/dynamoid/transaction/update.rb
ckhsponge: This was removed.

lib/dynamoid/transaction/update.rb
ckhsponge: This was removed.

lib/dynamoid/transaction/update.rb
ckhsponge: I don't mind the expression length in exchange for readability but I see the argument for brevity. I'll look into this.

lib/dynamoid/transaction/update.rb
ckhsponge: Thanks. I removed the puts.

lib/dynamoid/transaction/upsert.rb
ckhsponge: Yes! I changed update and delete to take an id which can be a hash_key or an [hash_key, range_key] like is used by find(). I don't think id is needed for create and upsert since those are both putting complete records. Should we include id anyways for those?

lib/dynamoid/transaction/executor.rb
ckhsponge: Removed this.

lib/dynamoid/transaction/create.rb
ckhsponge: changes_applied is now called on the model immediately after processing the model. This is earlier than the actual commit. We could wait and call changes_applied on all the models after the commit.

lib/dynamoid/transaction/update.rb
ckhsponge: Shorter update expressions are now implemented.

lib/dynamoid/transaction/create.rb
andrykonchin: Yeah, I believe `changes_applied` should be called only after successful `TransactWriteItems` call.

andrykonchin: Regarding `persisted?` - I haven't noticed where a new model is being marked as persisted.

lib/dynamoid/transaction/executor.rb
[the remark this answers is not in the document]
andrykonchin: Seems fine.

It makes sense also to have `after_commit` and `after_rollback` but it could be added later.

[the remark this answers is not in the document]
andrykonchin: > This is how ActiveRecord does it

The only difference - in case of Rails persisted changes are visible and can be read in any `after_` callback. But with the current approach in this PR any `after_` callback will not read the changes because they aren't still persisted. It may be confusing.

lib/dynamoid/transaction/delete.rb
```diff
@@ -0,0 +1,35 @@
...
+      def initialize(model_or_model_class, id = nil, options = {})
+        model = model_or_model_class.is_a?(Dynamoid::Document) ?
+                  model_or_model_class : model_or_model_class.find(id, consistent_read: true)
+        super(model, attributes, options)
+      end
+
+      def run_callbacks
+        model.run_callbacks(:destroy) do
+          yield if block_given?
+        end
```
andrykonchin, line 16: `#delete` in Rails doesn't run callbacks, only `#destroy`

lib/dynamoid/transaction/delete.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'action'
+
+module Dynamoid
+  class Transaction
+    class Delete < Action
+
+      def initialize(model_or_model_class, id = nil, options = {})
+        model = model_or_model_class.is_a?(Dynamoid::Document) ?
+                  model_or_model_class : model_or_model_class.find(id, consistent_read: true)
```
andrykonchin, line 9: **minor:** It seems loading model from the storage is not needed.

lib/dynamoid/transaction/update.rb
```diff
@@ -0,0 +1,69 @@
...
+        update_expression = options[:update_expression] || ("SET " + item_keys.each_with_index.map{ |k, i| "##{k} = :_s#{i}" }.join(", "))
+
+        # e.g. {":updated_at" => 1645453.234, ":i" => 1}
+        expression_attribute_values = item_keys.each_with_index.map{|k,i| [":_s#{i}",item[k]]}.to_h
+
+        if options[:add]
+          add_keys = options[:add].keys
+          update_expression += " ADD " + add_keys.each_with_index.map{|k,i| "#{k} :_a#{i}"}.join(", ")
+          add_keys.each_with_index{|k,i| expression_attribute_values[":_a#{i}"] = options[:add][k] }
+        end
```
andrykonchin, line 44: I would postpone supporting `update_expression` and `add` options until they are asked for.

lib/dynamoid/transaction/upsert.rb
```diff
@@ -0,0 +1,32 @@
...
+        changes = changes.merge attributes if attributes.present?
+        changes = add_timestamps(changes, skip_created_at: false) # overwrites created_at since this will be a new record essentially
+        changes[model_class.hash_key] ||= model.hash_key || SecureRandom.uuid
+        changes[model_class.range_key] ||= model.range_key if model_class.range_key? && model.range_key.present?
+        item = Dynamoid::Dumping.dump_attributes(changes, model_class.attributes)
+        {
+          put: {
+            item: sanitize_item(item),
+            table_name: model_class.table_name,
+          }
```
andrykonchin, line 27: It seems `put` that behaves like `PutItem` operation and just **replaces** existing item with provided attributes when item with primary key exists. But we need upserting - **partial updating** of existing item. So I suppose `update` operation should be used.

andrykonchin (commented):
Looks good!

lib/dynamoid/transaction/delete.rb
```diff
@@ -0,0 +1,35 @@
...
+          yield if block_given?
+        end
+      end
+
+      def to_h
+        key = {model_class.hash_key => model.hash_key}
+        key[model_class.range_key] = model.range_key if model_class.range_key?
+        # force fast fail i.e. won't retry if record is missing
+        condition_expression = "attribute_exists(#{model_class.hash_key})"
+        condition_expression += " and attribute_exists(#{model_class.range_key})" if model_class.range_key? # needed?
```
andrykonchin, line 24: Not sure we need to check whether item exists. If the condition fails the whole transaction will fails. I would like to have behaviour similar to Rails as mach as possible.

lib/dynamoid/transaction/action.rb
```diff
@@ -0,0 +1,71 @@
...
+        model.valid?
+      end
+
+      def raise_validation_error?
+        !!options[:raise_validation_error]
+      end
+
+      def run_callbacks
+        yield if block_given?
+      end
```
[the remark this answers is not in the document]
ckhsponge: An action isn't required to have callbacks. This is a base implementation that essentially does nothing and perhaps would be useful for a delete action or other actions that don't want to implement anything.

lib/dynamoid/transaction/delete.rb
ckhsponge: Indeed. I will rename this class to destroy and we can also have a delete method which is shorthand for destroy(skip_callbacks: true, skip_validation: true)

lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
```diff
@@ -8,6 +8,7 @@
 require_relative 'aws_sdk_v3/item_updater'
 require_relative 'aws_sdk_v3/table'
 require_relative 'aws_sdk_v3/until_past_table_status'
+require_relative 'aws_sdk_v3/transact'
```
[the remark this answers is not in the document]
ckhsponge: Any suggestions? This class wraps TransactWriteItems and someday maybe TransactGetItems as named by AWS.

lib/dynamoid/transaction/delete.rb
ckhsponge: Agreed. This is not needed for delete nor destroy.

lib/dynamoid/transaction/delete.rb
ckhsponge: I'm adding support for delete, destroy and destroy! The destroys will keep the existence check, delete will not.

lib/dynamoid/transaction.rb
```diff
@@ -0,0 +1,98 @@
...
+        yield(self)
+        commit
+      end
+    end
+
+    def self.execute(options = {})
+      transaction = new(options)
+      yield(transaction) if block_given?
+      transaction.commit
+    end
```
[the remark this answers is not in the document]
ckhsponge: Yes.

lib/dynamoid/transaction.rb
```diff
@@ -0,0 +1,98 @@
+
+require 'dynamoid/transaction/action'
+require 'dynamoid/transaction/create'
+require 'dynamoid/transaction/delete'
+require 'dynamoid/transaction/update'
+require 'dynamoid/transaction/upsert'
+
+module Dynamoid
+  class Transaction
```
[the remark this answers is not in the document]
ckhsponge: Renamed to TransactionWrite.

lib/dynamoid/transaction.rb
```diff
@@ -0,0 +1,98 @@
...
+    attr_accessor :action_inputs
+
+    def initialize(options = {})
+      @skip_callbacks = !!options[:skip_callbacks]
+      @skip_validation = !!options[:skip_validation]
```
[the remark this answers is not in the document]
ckhsponge: OK

ckhsponge:
@andrykonchin Thanks for the feedback. There are now proper delete and destroys. Upsert is working like an upsert as well.

lib/dynamoid/adapter_plugin/aws_sdk_v3.rb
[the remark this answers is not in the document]
andrykonchin: TBH I would move `transact_write_items` into the main plugin class. It's small and simple enough (what was a reason to move some methods into separate classes)

lib/dynamoid/transaction/delete.rb
andrykonchin: The main difference (in Rails) between `delete` and `destroy` is that the last one runs callbacks and needs instance. The first one doesn't need to have a model instantiated, but it loads a model now. This seems excessive.

andrykonchin:
Looks good so we can move on and and add specs.

lib/dynamoid/transaction/delete.rb
ckhsponge: delete no longer instantiates.

lib/dynamoid/transaction/create.rb
ckhsponge: new_record is now set to false after the commit.

I added an accessor for destroyed so that can be set when destroying.

ckhsponge:
@andrykonchin There's now a collection of specs for the transaction writes. Can you run them on github or give me permission to run them? 

What else is needed to get v1 released?

Thanks!

andrykonchin:
The new specs look good, but I would prefer to have more granular test cases that check only one aspect of a method behaviour (ideally only one check/assert)

ckhsponge:
> The new specs look good, but I would prefer to have more granular test cases that check only one aspect of a method behaviour (ideally only one check/assert per case)

Yeah, I usually put 4 or 5 syntax checks in each case. I'll split those out.

lib/dynamoid/transaction/upsert.rb
ckhsponge: @andrykonchin I assumed that the partition key could be changed but that is not true. A record must be deleted and recreated if you wish to change the partition key. How would you feel about simplifying the APIs for update and upsert?

Instead of:
`txn.update!(User, '1', {name: 'bob'} )`
it would be:
`txn.update!(User, {id: '1', name: 'bob'} )`

When there's a range key:
`txn.update!(User, ['1', 26], {name: 'bob'} )`
would be:
`txn.update!(User, {id: '1', age: 26, name: 'bob'} )`

This makes the API similar to create which I like but perhaps it's too inconsistent with the other update and upserts.

ckhsponge:
@andrykonchin Thanks for the specs review. What are your thoughts on this one?
https://github.com/Dynamoid/dynamoid/pull/688#discussion_r1436639724

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+
+    context 'simple primary key' do
+      it 'creates' do
+        obj1 = klass.new(name: "one")
+        obj2 = klass.new
+        obj3 = nil
+        expect(obj1.persisted?).to eql(false)
+        Dynamoid::TransactionWrite.execute do |txn|
+          txn.create! obj1
+          txn.create! obj2, {name: "two"}
```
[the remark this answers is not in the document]
ckhsponge: I agree that accepting additional attributes on the model-create is unlikely to be used. Since they will be used on the class-create and it's the same method it will work, however. Were you thinking we should put it extra effort and raise an exception to prevent it from working?

Allowing the extra attributes on all uses of update! does make sense I believe.

lib/dynamoid/transaction/upsert.rb
andrykonchin: > This makes the API similar to create which I like but perhaps it's too inconsistent with the other update and upserts.

It seems to me it will not work for updating a range key attribute. Its current value should be used to find an item and a new value to update.

spec/dynamoid/transaction_write_spec.rb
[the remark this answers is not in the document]
andrykonchin: I believe it's OK as is and checking actual arguments with raising exception or just warning could be a followup.

lib/dynamoid/transaction/upsert.rb
ckhsponge: No part of a primary key can be updated. The record must be deleted and re-created to change the key. If it's a composite key this includes the range key which cannot be changed.

I see your point about wanting to be consistent with the non-transaction methods however not allowing a separate key to be specified may be a better direction to go in. This will discourage people from trying to update primary keys among other things. I like the shorter method signature and that all the transaction method signatures are more similar as well.

andrykonchin:
A failing spec is fixed on master.

lib/dynamoid/transaction/upsert.rb
andrykonchin: > No part of a primary key can be updated. The record must be deleted and re-created to change the key. If it's a composite key this includes the range key which cannot be changed.

Indeed, exception is raised:
```
One or more parameter values were invalid: Cannot update attribute age. This attribute is part
 of the key (Aws::DynamoDB::Errors::ValidationException)
```

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+        field :name
+        validates :name, length: { minimum: 4 }
+      end
+    end
+
+    before do
+      klass.create_table
+      klass_with_composite_key.create_table
+      klass_with_callback.create_table
+      klass_with_validation.create_table
```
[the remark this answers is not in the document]
ckhsponge: I don't understand. Are you suggesting to move these create_table calls or maybe have less of them?

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+
+      it 'updates' do
+        obj1 = klass.create!(name: "one")
+        obj2 = klass.create!(name: "two")
+        obj3 = klass.create!(name: "three")
+        obj4 = klass.create!(name: "four")
+        Dynamoid::TransactionWrite.execute do |txn|
+          obj1.name = "oneone"
+          txn.update! obj1
+          txn.update! obj2, nil, {name: "twotwo"}
```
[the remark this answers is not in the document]
ckhsponge: Simplified the method signature so nil is not needed. There's more validation tests now that cover all combinations of arguments so hopefully we are good here.

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+        expect(obj3_found.name).to eql("threethree")
+        expect(obj4_found.name).to eql("four")
+      end
+
+      it 'upserts' do
+        obj1 = klass.create!(name: "one")
+        obj2 = klass.create!(name: "two")
+        obj3_id = SecureRandom.uuid
+        obj4_written = nil
+        obj5 = klass.new(id: SecureRandom.uuid, name: "five")
```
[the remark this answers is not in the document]
andrykonchin: Ah, this object is used for `upsert`.

There is no `upsert` instance method currently, only class method. So there shouldn't be available a method that manipulates an object in transaction interface. Or at least it shouldn't be tested.

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+    end
+
+    let(:klass_with_callback) do
+      new_class(class_name: 'Dog') do
+        field :name
+        field :says
+        field :says_more
+        before_save { self.says = "woof #{self.says_more}" }
+      end
+    end
```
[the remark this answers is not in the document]
ckhsponge: There are now tests for before and after for :save, :create, :update and :destroy callbacks. Order is checked.

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+        obj1_found = klass.find(obj1.id)
+        obj2_found = klass.find(obj2.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj2_found).to eql(obj2)
+        Dynamoid::TransactionWrite.execute do |txn|
+          txn.delete obj1
+          txn.delete klass, obj2.id
+        end
+        obj1_found = klass.find(obj1.id, raise_error: false)
+        obj2_found = klass.find(obj2.id, raise_error: false)
```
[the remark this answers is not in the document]
ckhsponge: Done.

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+        obj1_found = klass.find(obj1.id)
+        obj2_found = klass.find(obj2.id)
+        obj3_found = klass.find(obj3.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj2_found).to eql(obj2)
+        expect(obj3_found).to eql(obj3)
+        expect(obj1_found.name).to eql("one")
+        expect(obj2_found.name).to eql("two")
+        expect(obj3_found.name).to eql("three")
+      end
```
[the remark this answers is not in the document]
ckhsponge: Validation tests now check there is no rollback with non-bang saves and that there is a rollback with bang! saves.

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+        Dynamoid::TransactionWrite.execute do |txn|
+          expect(txn.save(obj1)).to be_present
+        end
+
+        obj1_found = klass_with_validation.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql("oneone")
+      end
+
+      it 'raises' do
```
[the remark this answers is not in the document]
ckhsponge: More validation tests cover this now.

spec/dynamoid/transaction_write_spec.rb
```diff
@@ -0,0 +1,414 @@
...
+        expect(obj1_found).to eql(obj1)
+        expect(obj2_found).to eql(obj2)
+        expect(obj1_found.name).to eql("oneone")
+        expect(obj2_found.name).to eql("two")
+        expect(obj1_found.says).to eql("woof bark1")
+        expect(obj2_found.says).to eql("woof bark2")
+      end
+    end
+
+    context 'validates' do
```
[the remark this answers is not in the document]
ckhsponge: More validation tests cover this now.

spec/dynamoid/transaction_write_spec.rb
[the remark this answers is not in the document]
andrykonchin: Yeah, I would suggest to create needed tables in test cases itself. 

In the tests bellow usually only one table is used. Table creation takes time so creating not used tables for every test case and then removing them may have significant effect.

ckhsponge:
@andrykonchin update() and upsert() no longer allow a key to be specified separate from the attributes. A key must always be included in the model or in the attributes. This simplifies the method signatures and makes them consistent with the other transaction actions.

I believe I covered all your feedback on the specs so another round of review would be good. Thanks!!!

spec/dynamoid/transaction_write_spec.rb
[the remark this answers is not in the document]
ckhsponge: Wow. That reduced running time by 66%.

spec/dynamoid/transaction_write_spec.rb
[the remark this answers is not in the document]
ckhsponge: That's reasonable. In the current implementation validations are performed on upsert when an instance is used which could be useful. There's not a complete picture of the final object, however. There's no mention of being able to use an instance in the upsert documentation.

I don't mind removing the instance tests.

spec/dynamoid/transaction_write/put_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+            obj2 = txn.create! klass_with_composite_key.new(id: obj1.id, age: 2, name: "two")
+          end
+          obj1_found = klass_with_composite_key.find(obj1.id, range_key: 1)
+          obj2_found = klass_with_composite_key.find(obj1.id, range_key: 2)
+          expect(obj1_found).to eql(obj1)
+          expect(obj2_found).to eql(obj2)
+          expect(obj1_found.name).to eql("oneone")
+          expect(obj2_found.name).to eql("two")
+        end
+      end
```
andrykonchin, line 91: I believe validation should be tested here as well

spec/dynamoid/transaction_write/upsert_spec.rb
```diff
@@ -0,0 +1,250 @@
...
+        obj3_id = SecureRandom.uuid
+        Dynamoid::TransactionWrite.execute do |txn|
+          txn.upsert! klass, {id: obj3_id, name: "threethree"}
+        end
+        obj3_found = klass.find(obj3_id)
+        expect(obj3_found.created_at).to be_nil
+        expect(obj3_found.updated_at.to_f).to be_within(1.seconds).of Time.now.to_f
+      end
+
+      context 'validation' do
```
andrykonchin, line 183: The `.upsert` method `doesn't run validations and callbacks`. So at least this behaviour shouldn't be tested.

spec/dynamoid/transaction_write_spec.rb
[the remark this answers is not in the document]
andrykonchin: Yeah, I see. So there are still cases to check that a transaction is rolled back when some of operations fails on DynamoDB side, e.g. deleting/updating not existing item, creation of item which primary key already exists.

The documentation states that:

> DynamoDB rejects the entire TransactWriteItems request if any of the following is true:
> - A condition in one of the condition expressions is not met.
> - An ongoing operation is in the process of updating the same item.
> - There is insufficient provisioned capacity for the transaction to be completed.
> - An item size becomes too large (bigger than 400 KB), a local secondary index (LSI) becomes too large, or a similar validation error occurs because of changes made by the transaction.
> - The aggregate size of the items in the transaction exceeds 4 MB.
> - There is a user error, such as an invalid data format.

So all mentioned by me cases are covered by the first documented scenario - `A condition in one of the condition expressions is not met`. So it makes sense to add a case for each transactional method (create/update/delete etc) to ensure that when a method logically fails on DynamoDB side - the whole transaction is rolled back. So transaction is atomic.

spec/dynamoid/transaction_write/update_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+    context 'updates' do
+      context 'simple primary key' do
+        before do
+          klass.create_table
+        end
+        it 'with attribute outside transaction' do
+          obj1 = klass.create!(name: "one")
+          obj1.name = "oneone"
+          Dynamoid::TransactionWrite.execute do |txn|
+            txn.update! obj1
```
andrykonchin, line 22: `update` with instance has different signature:

https://github.com/Dynamoid/dynamoid/blob/398f0657c68ef42aa41fa995f8a137ba6096d0ad/lib/dynamoid/persistence.rb#L748-L750

It's OK to skip this method for now.

spec/dynamoid/transaction_write/update_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+          expect(obj3_found).to eql(obj3)
+          expect(obj3_found.name).to eql("threethree")
+        end
+      end
+
+      it 'updates timestamps of instance' do
+        klass.create_table
+        obj1 = klass.new(name: "one", created_at: Time.now - 48.hours, updated_at: Time.now - 24.hours)
+        Dynamoid::TransactionWrite.execute do |txn|
+          txn.create! obj1
```
andrykonchin, line 87: Is it supposed to be the `create!` method call?

spec/dynamoid/transaction_write/upsert_spec.rb
andrykonchin: So there is no `upsert!` method (with `!`), only `upsert`.

andrykonchin (commented):
Specs look much better. Thank you!

spec/dynamoid/transaction_write/update_spec.rb
ckhsponge: OK, skipping support for blocks in update() for now. I would like to support increment eventually.

spec/dynamoid/transaction_write/put_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+# Dynamoid.config.logger.level = :debug
+
+describe Dynamoid::TransactionWrite do
+
+  describe '.execute' do
+    include_context 'transaction_write'
+
+    # a 'put' is a create that overwrites existing records if present
+    context 'puts' do
```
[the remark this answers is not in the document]
ckhsponge: I'm allowing create() to behave like a "put" when `skip_existence_check: true` is used. This does an overwrite if a previous record exists. The relatively short put_spec tests this. Maybe renaming to `allow_overwrite: true` would be better? Or add a put()?

Currently in a transaction when trying to create a record when it already exists raises Aws::DynamoDB::Errors::TransactionCanceledException.

spec/dynamoid/transaction_write/update_spec.rb
ckhsponge: Yes. Fixed.

spec/dynamoid/transaction_write/create_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+            Dynamoid::TransactionWrite.execute do |txn|
+              txn.create! obj2
+              txn.create! obj1
+            end
+          }.to raise_error(Dynamoid::Errors::DocumentNotValid)
+          expect(obj1.id).to be_present
+          expect(klass_with_validation.exists?(obj1.id)).to be_falsey
+          expect(obj2.id).to be_present
+          expect(klass_with_validation.exists?(obj2.id)).to be_falsey
+        end
```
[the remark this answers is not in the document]
ckhsponge: Added explicit rollback specs.

spec/dynamoid/transaction_write/save_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        it 'raises DocumentNotValid when not valid' do
+          obj1 = klass_with_validation.new(name: "one")
+          expect {
+            Dynamoid::TransactionWrite.execute do |txn|
+              txn.save! obj1
+            end
+          }.to raise_error(Dynamoid::Errors::DocumentNotValid)
+          expect(obj1.id).to be_present
+          expect(klass_with_validation.exists?(obj1.id)).to be_falsey
+        end
```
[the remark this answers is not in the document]
ckhsponge: Added explicit rollback specs.

spec/dynamoid/transaction_write/context.rb
```diff
@@ -0,0 +1,46 @@
...
+      before_save { self.says = "#{self.says_more} saving" }
+      after_save { self.says = "#{self.says} saved" }
+      before_create { self.says = "#{self.says} creating" }
+      after_create { self.says = "#{self.says} created" }
+      before_update { self.says = "#{self.says} updating" }
+      after_update { self.says = "#{self.says} updated" }
+      before_destroy { self.says = "#{self.says_more} destroying" }
+      after_destroy { self.says = "#{self.says} destroyed" }
+      # before_validate { self.says = "#{self.says} validating" }
+    end
```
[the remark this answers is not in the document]
ckhsponge: I couldn't get this to work immediately with the generated classes. Switched to using print and stdout checks like is done elsewhere.

spec/dynamoid/transaction_write/upsert_spec.rb
```diff
@@ -0,0 +1,250 @@
...
+        it 'with attribute outside transaction' do
+          obj1 = klass.create!(name: "one")
+          obj1.name = "oneone"
+          Dynamoid::TransactionWrite.execute do |txn|
+            txn.upsert! obj1
+          end
+          obj1_found = klass.find(obj1.id)
+          expect(obj1_found).to eql(obj1)
+          expect(obj1_found.name).to eql("oneone")
+        end
```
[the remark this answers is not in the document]
ckhsponge: upsert!() was removed and specs for upsert() with an instance were removed.

spec/dynamoid/transaction_write/upsert_spec.rb
```diff
@@ -0,0 +1,250 @@
...
+        it 'with class constructed in transaction' do
+          obj3_id = SecureRandom.uuid
+          Dynamoid::TransactionWrite.execute do |txn|
+            txn.upsert! klass, {id: obj3_id, name: "threethree"}
+          end
+          obj3_found = klass.find(obj3_id)
+          expect(obj3_found.id).to eql(obj3_id)
+          expect(obj3_found.name).to eql("threethree")
+        end
+        it 'with class instantiated outisde transaction' do
```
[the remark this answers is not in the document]
ckhsponge: Fixed.

spec/dynamoid/transaction_write/create_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+        klass_with_callback.create_table
+        obj2 = klass_with_callback.new(name: "two")
+        Dynamoid::TransactionWrite.execute do |txn|
+          txn.create! obj2, {says_more: "bark2"}
+        end
+        obj2_found = klass_with_callback.find(obj2.id)
+        expect(obj2_found).to eql(obj2)
+        expect(obj2_found.name).to eql("two")
+        expect(obj2_found.says).to eql("bark2 saving creating")
+        expect(obj2.says).to eql("bark2 saving creating created saved") # unpersisted changes
```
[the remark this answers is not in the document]
ckhsponge: print() and stdout are now used for these tests. I believe the order is checked by verifying the order of stdout.

Tests for around callbacks were added although I believe this is mostly adding coverage for the callback system as the transaction system has no logic for :before vs. :after vs. :around.

ckhsponge:
@andrykonchin More tests were added for rollbacks and callbacks. Instance usage tests for upsert were removed.

ckhsponge:
I removed the named parameter for options. This simplifies the brackets for expected usage e.g.:
```
save!( instance, skip_callbacks: true)
create!( klass, name: 'hi')
create!( klass, {name: 'hi'}, {skip_callbacks: true})
```
instead of:
```
save!( instance, options: {skip_callbacks: true})
create!( klass, {name: 'hi'})
create!( klass, {name: 'hi'}, options: {skip_callbacks: true})
```

ckhsponge:
@andrykonchin I believe all feedback has been addressed and further review would be appreciated.

ckhsponge:
@andrykonchin Anything blocking this? Thanks.

andrykonchin:
Could you please fix Rubocop issues (related to the changes only - there are some existing issues on master - I will address them myself) and squash all the working commits into one or more atomic ones?

andrykonchin:
I want to release a new version without transactions now and release transactions in the next version. The last release was a year ago and I was already asked for new version.

ckhsponge:
@andrykonchin I squashed the commits and I believe Rubocops will pass. Anything else I need to do? There are Codeclimate complexity issues.

andrykonchin:
Merged! :tada: Great work!

ckhsponge:
That's great news! Thanks for the support.

PR #691, opened 2023-10-17
Fix type annotation of arg `options`

kiharito:
Hello. Thank you for your maintenance.

This PR fixes the type annotation of the argument `options` in the method that sets range key.


andrykonchin:
Thank you!

PR #718, opened 2024-02-14
transaction write update blocks

ckhsponge:
transaction write supports blocks for updates to allow incrementing a field or adding to or deleting from a set


ckhsponge:
@andrykonchin Here is another transaction write PR. Sorry if I'm flooding you! Feel free to delay if you want to release the other stuff first.

ckhsponge:
@andrykonchin This improves the transactions that got merged a few weeks ago.

lib/dynamoid/transaction_write/action.rb
```diff
@@ -37,6 +41,25 @@ def range_key
...
+      def add(values)
+        additions.merge!(values)
+      end
+
+      # deletes a value or values from a set type or simply sets a field to nil
+      def delete(values)
+        if values.is_a?(Hash)
+          deletions.merge!(values)
+        else
+          set(values => nil)
```
andrykonchin, line 59: Currently `ItemUpdater#delete` without value means removing the whole attribute in an item, not just its nullifying. In terms of [UpdateExpression](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#API_UpdateItem_RequestSyntax) it's `REMOVE` operation.

README_transact.md
```diff
@@ -49,6 +49,39 @@ Dynamoid::TransactionWrite.execute do |txn|
...
+    u.add(friend_ids: [1, 2])
+    u.delete(child_names: ['bebe'])
+  end
+
+  # creates a new record with frequency 1 or increments the existing frequency
+  txn.upsert!(Color, id: 'red') do |u|
+    u.add(frequency: 1)
+  end
+end
+```
```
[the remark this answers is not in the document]
ckhsponge: I think supporting those operations could be very useful for upsert such as when maintaining counts. I believe some situations would require a possible race condition without it. Or at least an extra operation. It's essentially no extra code for transaction writing. I suppose I could remove the mention of it in the docs if you'd prefer.

lib/dynamoid/transaction_write/action.rb
ckhsponge: Indeed. I'll look into this. My understanding is REMOVE is functionally the same as setting to nil so there's no way to write a test distinguishing them.

README_transact.md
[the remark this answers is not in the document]
andrykonchin: Yeah, it makes sense. And I am considering unifying both `.upsert` and `#update` in the next major release.

But now there is a value in copying existing interface. And changing them synchronously later.

ckhsponge:
@andrykonchin I added the REMOVE and removed the docs for upsert blocks. I believe this covers all the feedback but please let me know if there's anything else. Thanks!

lib/dynamoid/transaction_write/action.rb
```diff
@@ -37,6 +42,30 @@ def range_key
...
+      end
+
+      # deletes a value or values from a set type or simply sets a field to nil
+      def delete(field_or_values)
+        if field_or_values.is_a?(Hash)
+          deletions.merge!(field_or_values)
+        else
+          remove(field_or_values)
+        end
+      end
```
andrykonchin, line 67: Currently there is no `remove` operation in `Document#update` method. `delete` removes an attribute if no value is specified and deletes collection elements otherwise. It has benefits to keep this abstraction and don't expose to user what DynamoDB parameters are used - legacy AttributeUpdates or more flexible UpdateExpression.

I propose to have only `#delete` method and don't expose `#remove` public method (and don't test it directly - only by calling the `#delete` method)

spec/dynamoid/transaction_write/create_spec.rb
```diff
@@ -196,5 +196,17 @@
...
+      klass.create_table
+      obj1 = nil
+      described_class.execute do |txn|
+        obj1 = txn.create! klass do |c|
+          c.set(name: 'one')
+        end
+      end
+      obj1_found = klass.find(obj1.id)
+      expect(obj1_found.name).to eql('one')
+    end
```
andrykonchin, line 210: As far as `Document.create` method doesn't support add/set/delete operations I wouldn't test this behaviour (even if transactional implementation accepts it) .

spec/dynamoid/transaction_write/update_spec.rb
```diff
@@ -224,6 +224,134 @@
...
+            u.set(name: 'oneone')
+            u.add(favorite_names: Set['charlie', 'dan']) # must be enumerable
+          end
+        end
+        obj1_found = klass.find(obj1.id)
+        expect(obj1_found).to eql(obj1)
+        expect(obj1_found.name).to eql('oneone')
+        expect(obj1_found.favorite_names).to eql(Set.new(%w[adam ben charlie dan]))
+      end
+    end
```
andrykonchin, line 297: It makes sense to have a separate `context` for the set operation.

spec/dynamoid/transaction_write/update_spec.rb
```diff
@@ -224,6 +224,134 @@
           end
         }.to output('saving updating updated saved ').to_stdout
       end
+
+      context 'adds' do
+        it 'to nil' do
```
andrykonchin, line 229: The test title doesn't describe the behaviour - I suppose if current value is `nil` then `0` is used.

spec/dynamoid/transaction_write/upsert_spec.rb
```diff
@@ -94,5 +94,65 @@
...
+          txn.upsert klass, { id: obj3_id, name: 'threethree' } do |u|
+            u.add(favorite_numbers: [2, 3])
+          end
+        end
+        obj3_found = klass.find(obj3_id)
+        expect(obj3_found.id).to eql(obj3_id)
+        expect(obj3_found.name).to eql('threethree')
+        expect(obj3_found.favorite_numbers).to eql(Set[1, 2, 3])
+      end
+    end
```
andrykonchin, line 156: If the `.update` method doesn't except a block - this shouldn't be tested as well.

README_transact.md
```diff
@@ -49,6 +49,33 @@ Dynamoid::TransactionWrite.execute do |txn|
 end
 ```
 
+### Update items
+An item can be updated by providing the hash key, range key if applicable, and the fields to update.
+Updating fields can also be done within a block using the `set()` method.
+To increment a numeric value or to add values to a set use `add()` within the block.
+Similarly a field can be removed (set to nil) or values can be removed from a set by using `delete()` in the block.
```
andrykonchin, line 56: `set to nil` seems incorrect now as far as `#delete` uses `REMOVE` that `Removes one or more attributes from an item.` (I suppose it meant removing key-value pair, not just assigning `nil`)

README_transact.md
```diff
@@ -49,6 +49,33 @@ Dynamoid::TransactionWrite.execute do |txn|
...
+Upsert does not officially support using blocks but documentation for this is coming soon.
```
andrykonchin, line 57:
```suggestion
```

lib/dynamoid/transaction_write/action.rb
ckhsponge: Agreed. I wasn't intending for remove() to be used publicly. I removed it.

spec/dynamoid/transaction_write/create_spec.rb
ckhsponge: Removed.

spec/dynamoid/transaction_write/update_spec.rb
ckhsponge: I added more contexts within 'add':
```
updates
...
      adds
        a value
          to an existing value
          to nil which defaults to zero
        to a set
          a set of strings
          an array
          a set of numbers
```

spec/dynamoid/transaction_write/update_spec.rb
ckhsponge: Clarified with 'to nil which defaults to zero'.

spec/dynamoid/transaction_write/upsert_spec.rb
ckhsponge: I think you meant `.upsert`. This is useful functionality that hopefully makes it back to the docs and tests soon. I removed these tests.

README_transact.md
ckhsponge: Removed.

ckhsponge:
@andrykonchin I clarified some contexts and removed some docs and tests for unsupported functionality. Let me know if I missed anything.

andrykonchin:
I will look at the changes a bit later.

spec/dynamoid/transaction_write/upsert_spec.rb
andrykonchin: Right 👍

spec/dynamoid/transaction_write/update_spec.rb
andrykonchin: I meant to test `add` and `set` operations separately, in different contexts and different test cases. Now they are tested in the same cases.

README_transact.md
ckhsponge: Removed "(set to nil)".

spec/dynamoid/transaction_write/update_spec.rb
ckhsponge: OK. I made a new context just for setting and also added 2 tests to ensure setting and adding can happen in the same block.
```
      sets and adds in a block
        to an existing value
        an array
      adds
        a value
          to nil which defaults to zero
          to an existing value
        to a set
          a set of numbers
          an array
          a set of strings
      sets in a block
        a string
```

ckhsponge:
@andrykonchin Just updated the PR with those issues.

andrykonchin:
Looks good. Thank you!

Issue #657, opened 2023-05-22
Conditional updates are incorrect in README


jplaisted:
I've discovered, that at least for `upsert`, the readme's examples of using conditions are incorrect. Namely; there is no `if:` keyword argument to that method. Instead, there is a positional argument at the end.

```ruby
# bad (from readme)
Address.upsert(id, { city: 'Chicago' }, if: { unless_exists: [:id] })
# good 
Address.upsert(id, { city: 'Chicago' }, { unless_exists: [:id] })
```

I see this `if` keyword elsewhere in the README and I sadly don't have time to track down all instances where it may be correct (or incorrect).

Something that may help here is, in addition to fixing the readme, adding more tests. I don't see `unless_exists` in any spec. Everything in the README aught to be in a test :)

## What happens if you run the bad example?

Dynamoid will never update items. This isn't a syntax error; I think its some ruby implicit kwargs to positional hash sillyness. Here's the definition of `upsert`

```ruby
def upsert(hash_key_value, range_key_value = nil, attrs = {}, conditions = {})
```

So what ends up happening is the `conditions` variable ends up as `{:if=>{:unless_exists=>[:id]}}`. The rest of the code never expects this `if`, and treats it literally. The update item call to dynamo will have a `expected:{"unless_exists"=>{value:{l:[{s:"id"}]}}}`, which is not correct, and will cause a failure condition.

## What happens if you run the good example?

Things work as intended. Code [here](https://github.com/Dynamoid/dynamoid/blob/aa790f15498bbdc1c6108fff8b80b04630b8c27b/lib/dynamoid/adapter_plugin/aws_sdk_v3.rb#LL631C9-L633C12) knows to extract `unless_exists` from the conditions hash, and use that to form the dynamo call, which now will have `expected:{"id"=>{exists:false}}` instead.

andrykonchin:
Nice catch! Will fix README soon.

jplaisted:
Thanks!

PR #720, opened 2024-02-20
Incorrect code sample of update method in readme.md

ndjndj:
Hello, thank you for maintenance of this gem.

Probably related to issue #657.

I found what incorrect code sample of update method in readme.md.

`.update` method does not accept conditions as arguments.   

https://github.com/Dynamoid/dynamoid/blob/6b96d1d9cb8cb6d011cc84d1ca4cc65f43d6a729/lib/dynamoid/persistence.rb#L232

Therefore, the following code as shown in the sample will fail.

```
Address.update(id, { city: 'Chicago' }, if: { deliverable: true })
```

So I removed this code sample.

thank you!


andrykonchin:
Thank you! Could you please rebase the branch on master?

ndjndj:
I’m sorry. Is this correct?

andrykonchin:
Could you please squash all commits into one? I would do it myself and push with force into your branch but it seems it's disallowed.

PR #757, opened 2024-06-06
Allow ActiveSupport::HashWithIndifferentAccess of yaml parse

llpereiras:
The serialized field are stored with ActiveSupport

In my model, I added this:

```ruby
  field :description, :serialized
```

image

when I read the record, I received this error:

```
Psych::DisallowedClass: Tried to load unspecified class: ActiveSupport::HashWithIndifferentAccess
```

I think, the most of the apps use rails and ActiveSupport, so I think should be a good addiction to add the support of ActiveSupport

### DIFF
--- a/lib/dynamoid/undumping.rb
+++ b/lib/dynamoid/undumping.rb
@@ -256,7 +256,9 @@ module Dynamoid
           # String
           # Array
           # Hash
-          YAML.safe_load(value, permitted_classes: [Symbol, Set, Date, Time, DateTime])
+          YAML.safe_load(value, permitted_classes: [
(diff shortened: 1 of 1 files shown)


lmpereirayouse (approved):

andrykonchin:
I agree that it would be convenient to allow serialising such classes like `HashWithIndifferentAccess`. So the only missing thing in the PR is specs for the new feature. Could you please add them?

But TBH I like mentioned above config option `yaml_column_permitted_classes`. And `use_yaml_unsafe_load` could also be useful or at least convenient for developers that are familiar with `ActiveRecord`. So I am open to adding them.

Issue #568, opened 2022-07-19
Single Table Design: Support Patterns


bmalinconico:
Hello,

I wanted to have a conversation as to how, and if at all, Dynamoid supports the common [single table design](https://aws.amazon.com/blogs/compute/creating-a-single-table-design-with-amazon-dynamodb/) for DynamoDB.

The basic TL;DR; is the existing STI with some Range Key magic. The Hash Key represents the ability to get all of a common thing, I'm going to use Pizza with a kinda contrived example.

| PizzaId(HK) | ResourceId (RK)  | type (Required for Deserialization) | Name     | Cost | Quantity | LeftHalf | RightHalf |
| --          | --               | --                                  | --       | --   | --       | --       | --        |
| a           | pizza            | Pizza                               | Combo    | 13.99
| a           | topping-onions   | PizzaTopping                        | Onion    |      | Heavy    | true     | true
| a           | topping-peperoni | PizzaTopping                        | Peperoni |      | Light    | false    | true

This breaks down with the Dynamoid STI implementation.

I'd like to open a conversation about updating the ORM to allow for this pattern and want to toss around some options then identify what would need to be updated to work.

You can query for all toppings on a pizza by using Pizza ID and RK Prefix

## First Idea
Allow a range prefix to be specified

```
class PizzaTopping
  range :resource_id, prefix_on_persistance: "topping-"
end

# There is also a need to "fix" the RK for single instance objects
class Pizza
  range :resource_id, fixed_value: "pizza"
end
```

andrykonchin:
Hi,

In short - Dynamoid doesn't support explicitly anything related to the Single-Table design. Why? I suppose because its goal is to implement the ActiveRecord pattern.

It seems to me that the classic ActiveRecord pattern contradicts with ideas of the Single-Table design with kind of schemeless/multi-schema items. But I can easily imagine that Single-Table design's approach may be implemented on top of ORM like Dynamoid. Or on top of any other DynamoDB client.

That is I am not against supporting Single-Table design in Dynamoid. But I see benefits in separating new features and existing ActiveRecord-like approach. How strong should be this separation? I don't know right now. It depends probably on how natural specific features look from the point of view of the ActiveRecord approach.

Regarding the proposed feature with a range prefix. I am not familiar with the concrete patterns of the Single-Table design and don't know whether such range prefixes is a common/well known pattern. Could you please point at resources that describe such patterns?

thomaswitt:
[...]

PR #822, opened 2024-11-19
Support binary type natively

dalibor:
aws-sdk-dynamodb [expects](https://github.com/aws/aws-sdk-ruby/blob/46208f4875a38b55cf47c73a03a8743c5fb4d57f/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/simple_attributes.rb#L26-L45) `StringIO` object for binary and [returns back](https://github.com/aws/aws-sdk-ruby/blob/46208f4875a38b55cf47c73a03a8743c5fb4d57f/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/attribute_value.rb#L88) the same object type for binary values.

Current implementation incorrectly converts and stores binary values as strings that due to Base64 encoding use 33% more storage and read/write capacity.

Request Before:

```
{
  "TableName": "dynamoid_tests_documents_1732010637_594",
  "Item": {
    "image": {
      "S": "AIj/"
    },
    ...
}
```

Request After:

```
{
  "TableName": "dynamoid_tests_documents_1732010679_801",
  "Item": {
    "image": {
      "B": "AIj/"
    },
    ...
}
```

### DIFF
--- a/lib/dynamoid/config.rb
+++ b/lib/dynamoid/config.rb
@@ -50,6 +50,7 @@ module Dynamoid
     option :store_date_as_string, default: false # store Date fields in ISO 8601 string format
     option :store_empty_string_as_nil, default: true # store attribute's empty String value as null
     option :store_boolean_as_native, default: true
+    option :store_binary_as_native, default: false
     option :backoff, default: nil # callable object to handle exceeding of table throughput limit
     option :backoff_strategies, default: {
       constant: BackoffStrategies::ConstantBackoff,
--- a/lib/dynamoid/dumping.rb
+++ b/lib/dynamoid/dumping.rb
@@ -297,10 +297,24 @@ module Dynamoid
       end
     end
 
-    # string -> string
+    # string -> StringIO
     class BinaryDumper < Base
       def process(value)
-        Base64.strict_encode64(value)
+        store_as_binary = if @options[:store_as_native_binary].nil?
+                            Dynamoid.config.store_binary_as_native
+                          else
+                            @options[:store_as_native_binary]
+                          end
+
+        if store_as_binary
+          if value.is_a?(StringIO) || value.is_a?(IO)
+            value
+          else
+            StringIO.new(value)
+          end
+        else
+          Base64.strict_encode64(value)
+        end
       end
     end
 
--- a/lib/dynamoid/type_casting.rb
+++ b/lib/dynamoid/type_casting.rb
@@ -288,7 +288,9 @@ module Dynamoid
 
     class BinaryTypeCaster < Base
       def process(value)
-        if value.is_a? String
+        if value.is_a?(StringIO) || value.is_a?(IO)
+          value
+        elsif value.is_a?(String)
           value.dup
         else
           value.to_s
--- a/lib/dynamoid/undumping.rb
+++ b/lib/dynamoid/undumping.rb
@@ -284,7 +284,17 @@ module Dynamoid
 
     class BinaryUndumper < Base
       def process(value)
-        Base64.strict_decode64(value)
+        store_as_binary = if @options[:store_as_native_binary].nil?
+                            Dynamoid.config.store_binary_as_native
+                          else
+                            @options[:store_as_native_binary]
+                          end
+
+        if store_as_binary
+          value.string # expect StringIO here
+        else
+          Base64.strict_decode64(value)
+        end
       end
     end
 
--- a/spec/dynamoid/dumping_spec.rb
+++ b/spec/dynamoid/dumping_spec.rb
@@ -1606,20 +1606,87 @@ describe 'Dumping' do
   end
 
   describe 'Binary field' do
-    let(:klass) do
-      new_class do
-        field :image, :binary
-      end
-    end
-
     let(:unfrozen_string) { +"\x00\x88\xFF" }
     let(:binary_value) { unfrozen_string.force_encoding('ASCII-8BIT') }
 
-    it 'encodes a string in base64-encoded format' do
-      obj = klass.create(image: binary_value)
+    context 'default non-native binary' do
+      let(:klass) do
+        new_class do
+          field :image, :binary
+        end
+      end
 
-      expect(reload(obj).image).to eql(binary_value)
-      expect(raw_attributes(obj)[:image]).to eql(Base64.strict_encode64(binary_value))
+      it 'encodes a string in base64-encoded format' do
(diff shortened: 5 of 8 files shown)


lib/dynamoid/undumping.rb
```diff
@@ -284,7 +284,7 @@ def process(value)
 
     class BinaryUndumper < Base
       def process(value)
-        Base64.strict_decode64(value)
+        value.read
```
andrykonchin, line 287: It seems `StringIO#read` makes a copy of a string (`#read` accepts a length parameter so in general case it can return a substring what is left to "read"), so `StringIO#string` looks slightly better from the performance/memory footprint point of view.

spec/dynamoid/dumping_spec.rb
```diff
@@ -1619,7 +1619,8 @@ def self.dynamoid_field_type
       obj = klass.create(image: binary_value)
 
       expect(reload(obj).image).to eql(binary_value)
-      expect(raw_attributes(obj)[:image]).to eql(Base64.strict_encode64(binary_value))
+      expect(raw_attributes(obj)[:image].class).to eql(StringIO)
+      expect(raw_attributes(obj)[:image].read).to eql(binary_value)
```
andrykonchin, line 1623: Probably the test title is not accurate anymore.

andrykonchin (commented):
Looks good! 

Could you please take a look on the comments?

spec/dynamoid/dumping_spec.rb
andrykonchin: Another concern is String encoding. When document is loaded with manual `Base64.strict_encode64` call result has US-ASCII encoding. We need to ensure it isn't changed with this change. Probably we need new tests for this.

andrykonchin:
Started a discussion https://github.com/aws/aws-sdk-ruby/discussions/3146.

lib/dynamoid/dumping.rb
```diff
@@ -297,10 +297,14 @@ def process(value)
...
-    # string -> string
+    # string -> StringIO
     class BinaryDumper < Base
       def process(value)
-        Base64.strict_encode64(value)
+        if value.is_a?(StringIO) || value.is_a?(IO)
+          value
+        else
+          StringIO.new(value)
+        end
```
andrykonchin, line 307: I suppose that now new values of a binary field will be stored in DynamoDB as `binary` but all existing values are stored as `string`. It might be a bit surprising for end users.

The common approach is to steadily migrate users from the current behaviour to the new one. We can introduce a new field's option to either keep old behaviour (by default, so there is no a breaking change in a minor release) or to switch to the new one (for new fields or new tables or in new projects).

Example is a `boolean` type that wasn't supported by DynamoDB initially so Dynamoid's `boolean` fields were stored as `f` and `t`. Later DynamoDB added support of the boolean type and a new option `store_as_native_boolean` was added:

```ruby
field :active, :boolean, store_as_native_boolean: false
```

Later in the next major release such migration options can be removed and we can propose users to  custom types to keep old storage formats.

So I propose to add a new field's option `store_as_native_binary` that is by default false. And check it at casting, dumping and undumping.

lib/dynamoid/config.rb
```diff
@@ -50,6 +50,7 @@ module Config
     option :store_date_as_string, default: false # store Date fields in ISO 8601 string format
     option :store_empty_string_as_nil, default: true # store attribute's empty String value as null
     option :store_boolean_as_native, default: true
+    option :store_binary_as_native, default: false
```
andrykonchin, line 53: Yeah, it makes sense to have also a global config option. But it doesn't affect field conversion and isn't tested yet.

lib/dynamoid/type_casting.rb
```diff
@@ -288,8 +288,8 @@ def process(value)
 
     class BinaryTypeCaster < Base
       def process(value)
-        if value.is_a? String
-          value.dup
+        if value.is_a?(StringIO) || value.is_a?(IO)
+          value
         else
           value.to_s
         end
```
andrykonchin, line 295: Not sure if it's correct to remove the `#dup` call. Almost all parameters are duplicated at casting. Don't remember the reason and cannot find quickly a relevant GibHub issue. But if duplication should be removed - it should be done for all the field types in this file.

lib/dynamoid/dumping.rb
dalibor: I agree. It could cause issues for integrators even though they might not have noticed that the binary values aren't stored with binary type in DynamoDB.

I've added the `store_as_native_binary` option that should be a safe work-around, but there's still a risk with this change to not get noticed with the new major release since there will not be a breaking code change.

This concerns me along with the lack of alignment with DynamoDB's single table design which is why I've started exploring `Aws::DynamoDB::Client` directly for my project.

Let me know if you want me to do any other changes to this PR, or I'll leave it to you for how you want to get it out.

lib/dynamoid/type_casting.rb
dalibor: Yeah, I noticed that too. Let me add it back just to be on the safe side. I'm not familiar with Dynamoid, just playing with it in the last couple days.

lib/dynamoid/dumping.rb
andrykonchin: The only way to prevent breaking changes being unnoticed that I see is to add deprecation warnings in the last minor release before the major one.

dalibor: Could also require the `store_as_native_binary` option be set, otherwise raise an error.

andrykonchin: Regarding the single-table design - it seems to me it's out of scope of Dynamoid. Or at least of the initial scope to implement the Active Record pattern.

I am not opposed to including the new functionality into the gem related to this "pattern". But I haven't found any common and widely excepted "interface" or API. So I postponed this.

There is an issue #568 but all I heard there were either some generic thoughts or requesting somebody's project specific features. So there was nothing I can work with.

andrykonchin (approved):
Looks great to me! Thank you.

andrykonchin:
> Code coverage seems to have already been broken in master. It might need a later ruby version?

Yeah, I didn't have a chance to fix this annoying issue. Please don't bother.

The only thing left is to squash working commits. I can do it on my own today later.

dalibor:
Cool - feel free to squash and merge when you'd like. Thank you!

PR #846, opened 2025-01-13
TransactionWrite update() with a block

ckhsponge:
WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ DynamoDB is not like other document-based databases you might know, and
 is very different indeed from relational databases. It sacrifices
 anything beyond the simplest relational queries and transactional
 support to provide a fast, cost-efficient, and highly durable storage
-solution. If your database requires complicated relational queries and
-transaction support, then this modest Gem cannot provide them for you,
+solution. If your database requires complicated relational queries
+then this modest Gem cannot provide them for you
 and neither can DynamoDB. In those cases you would do better to look
 elsewhere for your database needs.
 
@@ -1187,6 +1187,19 @@ Dynamoid::TransactionWrite.execute do |txn|
   # sets the name and title for a user
   # The user is found by id (that equals 1)
   txn.update_fields(User, '1', name: 'bob', title: 'mister')
+
+  # sets the name, increments a count and deletes a field
+  txn.update_fields(User, 1) do |u1| # a User instance is provided
+    u1.set(name: 'bob')
+    u1.add(article_count: 1)
+    u1.delete(:title)
+  end
+
+  # adds to a set of integers and deletes from a set of strings
+  txn.update_fields(User, 2) do |u2|
+    u2.add(friend_ids: [1, 2])
+    u2.delete(child_names: ['bebe'])
+  end
 end
 ```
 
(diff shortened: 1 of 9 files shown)


lib/dynamoid/transaction_write.rb
```diff
@@ -394,6 +395,27 @@ def update_attributes!(model, attributes)
...
+    #
+    # Returns +true+ if saving is successful and +false+
+    # otherwise.
+    #
+    # @param model [Dynamoid::Document] a model
+    # @return [true|false] Whether updating successful or not
+    def update(model, &block)
+      action = Dynamoid::TransactionWrite::Update.new(model, raise_error: false, &block)
+      register_action action
+    end
```
andrykonchin, line 417: This new transactional method `#update` behaves a bit differently than the non-transactional one. Non-transactional updates a model and its fields with new values. It's easy enough because `UpdateItem` may be configured to return all attributes of an updated item. `TransactWriteItems` doesn't support this AFAIK. And manually implement logic of `remove`/`add` operations to update attributes in memory also doesn't appeal to me. So accepting a model as a parameter seems incorrect to me.

Instead I would propose to accept a primary key as a parameter. There is no such non-transactional method but I will add it later to complement existing `#update` method.

lib/dynamoid/transaction_write/save.rb
```diff
@@ -86,6 +92,30 @@ def action_request
...
+      end
+
+      # deletes a value or values from a set type or simply sets a field to nil
+      def delete(field_or_values)
+        if field_or_values.is_a?(Hash)
+          @deletions.merge!(field_or_values)
+        else
+          remove(field_or_values)
+        end
+      end
```
andrykonchin, line 117: It makes sense to have a separate class to collect user actions (I mean to set/add/delete), similar to `ItemUpdater` (https://github.com/Dynamoid/dynamoid/blob/master/lib/dynamoid/adapter_plugin/aws_sdk_v3/item_updater.rb)

README.md
```diff
@@ -1106,6 +1106,7 @@ model
 * `#destroy`/`#destroy!` - remove an model
 * `#upsert` - add a new model or update an existing one, no callbacks
 * `#update_fields` - update a model without its instantiation
+* `#update` - update a model using a block that supports set, add, and delete operations on fields
```
andrykonchin, line 1109:
```suggestion
* `#update` - update a model using a block that supports set, add, remove, and delete operations on fields
```

lib/dynamoid/transaction_write.rb
ckhsponge: You are correct. Because of that should we just change update_fields to optionally allow a block or do you think adding an entirely new method to TransactionWrite would be better?

andrykonchin: I would prefer adding new method `update` in the `TransactionWrite` module (like it was already done here).

ckhsponge:
@andrykonchin I moved the updating-with-block code to update_fields and put the handling into an ItemUpdater class. Let me know if this looks good and I'll put the tests in for this.

andrykonchin:
Yeah, looks good!

andrykonchin:
Fixed the issue with `uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger` error on CI so CI should be green.

ckhsponge:
@andrykonchin The code and tests are back and checks are passing. The code climate check made me split things up a bit. Let me know how it looks.

andrykonchin:
Thank you! Will review and polish later next week.

ckhsponge:
Great!

ckhsponge:
My pleasure!

PR #889, opened 2025-03-21
Add YAML loading options to mirror ActiveRecord behavior

mlarraz:
Based on discussion in https://github.com/Dynamoid/dynamoid/pull/757, adds two new options that allow YAML safe loading behavior to be customized or disabled.

Because there is a lot of different behavior between versions of the Pysch gem, we should directly check what methods are available instead of the Ruby version (particularly because newer Ruby versions can still run older versions of Pysch).


andrykonchin:
Thank you! Looks good. Could you add specs for the new config options?

It seems to me the proper place for new tests is https://github.com/Dynamoid/dynamoid/blob/master/spec/dynamoid/dumping_spec.rb.

PR #897, opened 2025-04-18
TransactionWrite sanitizes items

ckhsponge:
TransactionWrite should call sanitize_item so that nils are not persisted. Otherwise GSI will fail with a nil range key.

### DIFF
--- a/lib/dynamoid/transaction_write/save.rb
+++ b/lib/dynamoid/transaction_write/save.rb
@@ -97,6 +97,7 @@ module Dynamoid
         touch_model_timestamps(skip_created_at: false)
 
         attributes_dumped = Dynamoid::Dumping.dump_attributes(@model.attributes, @model_class.attributes)
+        attributes_dumped = sanitize_item(attributes_dumped)
 
         # require primary key not to exist yet
         condition = "attribute_not_exists(#{@model_class.hash_key})"
@@ -119,6 +120,7 @@ module Dynamoid
         # changed attributes to persist
         changes = @model.attributes.slice(*@model.changed.map(&:to_sym))
         changes_dumped = Dynamoid::Dumping.dump_attributes(changes, @model_class.attributes)
+        changes_dumped = sanitize_item(changes_dumped)
 
         # primary key to look up an item to update
         key = { @model_class.hash_key => @model.hash_key }
--- a/spec/dynamoid/transaction_write/indexes_spec.rb
+++ b/spec/dynamoid/transaction_write/indexes_spec.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Dynamoid::Indexes do
+  let(:klass_with_gsi) do
+    new_class(class_name: 'Ferret') do
+      field :name
+      field :age, :number
+      global_secondary_index hash_key: :name, range_key: :age, projected_attributes: :all, name: "str_num"
+    end
+  end
+
+  before do
+    klass_with_gsi.create_table
+  end
+
+  describe '.global_secondary_index' do
+
+    it 'can save in transaction' do
+      doc = klass_with_gsi.new(name: 'abc', age: 1)
+      Dynamoid::TransactionWrite.execute do |txn|
+        txn.save! doc
+      end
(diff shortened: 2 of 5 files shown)


spec/dynamoid/transaction_write/indexes_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      doc.name = 'abc'
+      doc.age = 1
+      doc.save!
+      Dynamoid::TransactionWrite.execute do |txn|
+        txn.upsert klass_with_gsi, doc.id, age: nil
+      end
+      expect(doc).to be_persisted
+    end
+  end
+end
```
andrykonchin, line 71: It's kind of integration tests, so it makes sense to have them, but it will be great to have more thorough specs for each transactional method, like it's done for non-transactional ones:

https://github.com/Dynamoid/dynamoid/blob/master/spec/dynamoid/persistence_spec.rb#L2996-L3037

Such specs would test:
- what actual value of a nullified attribute is persisted (it either has `null` value or is removed in the item)
- and how `store_attribute_with_nil_value` option affects this behaviour

lib/dynamoid/transaction_write/save.rb
```diff
@@ -119,6 +120,7 @@ def action_request_to_update
         # changed attributes to persist
         changes = @model.attributes.slice(*@model.changed.map(&:to_sym))
         changes_dumped = Dynamoid::Dumping.dump_attributes(changes, @model_class.attributes)
+        changes_dumped = sanitize_item(changes_dumped)
```
andrykonchin, line 123: It seems it's more difficult than just skip setting `nil` values. A nullified attribute of an existing item simply will not be updated at all (so old value will be kept unchanged). So correct logic is to remove such attribute at all.

It's done in the following way in  non-transactional method `#save`:
- https://github.com/Dynamoid/dynamoid/blob/master/lib/dynamoid/persistence/save.rb#L44
- https://github.com/Dynamoid/dynamoid/blob/master/lib/dynamoid/adapter_plugin/aws_sdk_v3/item_updater.rb#L52-L66

ckhsponge: Indeed. The code I put in there just skips the nils which is not correct. I'll have to do the removing for all these methods like you said.

spec/dynamoid/transaction_write/indexes_spec.rb
ckhsponge: I'm thinking to not add tests for store_attribute_with_nil_value = true when there is an index. This fails even without transactions. I suppose I could add a test that a ValidationException is thrown in this case. And that test should exist for both transactions and non-transactions.

```
  let(:klass_with_gsi) do
    new_class(class_name: 'Ferret') do
      field :name
      field :age, :number
      global_secondary_index hash_key: :name, range_key: :age, projected_attributes: :all, name: 'str_num'
    end
  end
...
    context 'store_attribute_with_nil_value = true', config: { store_attribute_with_nil_value: true } do
      it 'can set gsi field to nil' do
        doc = klass_with_gsi.new
        doc.name = 'abc'
        doc.age = 1
        doc.save!
        doc.age = nil
        doc.save! #      Aws::DynamoDB::Errors::ValidationException: Invalid attribute value type
      end
    end
```

andrykonchin:
I'll review this later this week.

spec/dynamoid/transaction_write/indexes_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      doc.name = 'abc'
+      doc.age = 1
+      doc.save!
+      doc.age = nil
+      Dynamoid::TransactionWrite.execute do |txn|
+        txn.save! doc
+      end
+      expect(doc).to be_persisted
+      expect(doc.reload.age).to be_nil
+    end
```
andrykonchin, line 60: It makes sense to have similar specs for `#update_attributes` and `#update_attribute`.

ckhsponge: I added update_attributes, update_fields and update_fields with set().

andrykonchin (approved):
Looks good!

PR #910, opened 2025-06-01
Add key_type option to table method

ogidow:
My project uses `integer` type hash key.

Currently, if we specify integer key to hash key, we need to do below process on dynamoid
1. define hash key with `table` method
2. override hash key type by `field` method

```ruby
class Model
  include Dynamoid::Document
  
  table key: :hash_key # hash_key define as string type

  field :hash_key, :integer # override hash_key type to integer from string
end
```

But, when we override attribute, dynamoid notice us warning message like 
```
'Method hash_key= generated for the field hash_key overrides already existing method'
'Method hash_key generated for the field hash_key overrides already existing method'
``` 

So, I'd like to add key_type option to specify type to hash key on table method to avoid warning message.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,8 +7,10 @@ inherit_from:
   - .rubocop_todo.yml
 
 require:
-  - rubocop-md
   - rubocop-packaging
+
+plugins:
+  - rubocop-md
   - rubocop-performance
   - rubocop-rake
   - rubocop-rspec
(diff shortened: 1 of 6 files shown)


andrykonchin:
Thank you! Looks good. I was considering this feature for a long time.

andrykonchin:
There is nothing to do for you - I will handle the CI failure.

andrykonchin (approved):

PR #916, opened 2025-07-02
add config to error on scan

aaronalberg:
Adds an optional configuration option to raise an error on scan. Prevents accidental expensive/slow scans in situations when only queries are desired

### DIFF
--- a/lib/dynamoid/config.rb
+++ b/lib/dynamoid/config.rb
@@ -36,6 +36,7 @@ module Dynamoid
     option :read_capacity, default: 100
     option :write_capacity, default: 20
     option :warn_on_scan, default: true
+    option :error_on_scan, default: false
     option :endpoint, default: nil
     option :identity_map, default: false
     option :timestamps, default: true
--- a/lib/dynamoid/criteria/chain.rb
+++ b/lib/dynamoid/criteria/chain.rb
@@ -563,7 +563,7 @@ module Dynamoid
         if @key_fields_detector.key_present?
           raw_pages_via_query
         else
-          issue_scan_warning if Dynamoid::Config.warn_on_scan && !@where_conditions.empty?
+          validate_scan_conditions
           raw_pages_via_scan
         end
       end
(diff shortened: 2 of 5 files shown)


lib/dynamoid/criteria/chain.rb
```diff
@@ -598,6 +598,11 @@ def raw_pages_via_scan
         end
       end
 
+      def validate_scan_conditions
```
aaronalberg, line 601: code climate was complaining about cognitive complexity so I moved this portion of the method out

lib/dynamoid/config.rb
```diff
@@ -36,6 +36,7 @@ module Config
     option :read_capacity, default: 100
     option :write_capacity, default: 20
     option :warn_on_scan, default: true
+    option :error_on_scan, default: false
```
andrykonchin, line 39: Could you please also update the README.md file and document this config option?

https://github.com/Dynamoid/dynamoid/blob/master/README.md?plain=1#L1274

spec/dynamoid/criteria_spec.rb
```diff
@@ -133,4 +150,19 @@
...
+      Dynamoid::Config.error_on_scan = true
+    end
+
+    after do
+      Dynamoid::Config.error_on_scan = @error_on_scan
+    end
+
+    it 'does not log any warnings' do
+      expect { User.all }.not_to raise_error(Dynamoid::Errors::ScanProhibited)
+    end
```
andrykonchin, line 166: Typo in `does not log any warnings`.

andrykonchin (commented):
Looks good. Could you please squash commits into a single one?

spec/dynamoid/criteria_spec.rb
aaronalberg: ah yeah thank you! fixed

lib/dynamoid/config.rb
aaronalberg: updated

aaronalberg:
@andrykonchin I noticed that releases have been happening only once a year. Does this mean that I'll need to wait until next year to see this change released? Or is there a way to get this out sooner?

andrykonchin (approved):

